PR #206, opened 2015-01-13
Fix background colors for github enterprise (Fixes #205)

### LINKED ISSUE #205: Background issues on GitHub Enterprise
On GHE, the diff backgrounds do not extend through the code. This makes it difficult to see where the diff lines are. The cause is an extra ```<pre>``` tag that needs to be styled. PR coming shortly.

chimericdream:


Mottie:
Thanks for contributing! :+1:

PR #233, opened 2015-05-08
Add font size option

Poorchop:
I don't know how user options work with userstyles but I'm just guessing that this is the proper syntax based on the existing CSS. A font size option would be great because my font of choice looks awful at the default size. Let me know if I need to do anything else in order for this option to appear on the userstyles website. I forgot to make a branch in my local repo but oh well.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -17,6 +17,7 @@
   pre, code, tt, kbd:not(.badmono), pre, samp, .blob-code, .file-data pre, .line-data, #gist-form .file .input textarea,
   .ace_editor.ace-github-light, .ace_editor {
       font-family: "/*[[font-choice]]*/", Monaco, Consolas, "Liberation Mono", Menlo, Courier, monospace !important;
+      font-size: /*[[fontsize-choice]]*/px !important;
   }
   /* Base link/panel colors - text */
   a:not([href*="/labels/"]), .profilecols ul.stats li a:hover strong, .plans-row .plan, .filters li.selected, .vcard-stat,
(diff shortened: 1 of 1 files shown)


silverwind:
I'd rename that option to just `font-size`. The rest is done through the userstyles.org interface.

Poorchop:
All of the other options had different names for the properties that they were affecting, so I wasn't sure if there would be a conflict naming the option `font-size`.

silverwind:
This could potentially break layout, as I'm not sure Github uses the same font-size everywhere.

Ideally, a default value shouldn't set a font-size rule for compatbilty, so an 'default' option would be necessary. This raises the issue of not being able to set a non-changing default, as one can't just `inherit` the font-size. Hard nut to crack ;)

Mottie:
Actually, it shouldn't be difficult since the userstyles site allows nesting of placeholders. So you could have an option in the github-dark.css file like this:

```css
/*[[add-font-size]]*/
```

then on the userstyles site, have that be replaced by this definition if a font-size is chosen:

```css
font-size: /*[[font-size]]*/px !important;
```

then whatever the user selects will replace the `/*[[font-size]]*/` placeholder.

Mottie:
Oh, you only want to change the code font size? I think we'd also have to change the line-height, which I think is set in pixels... so yeah, this could get messy.

Poorchop:
Looks like there might be some consistency issues and I guess this is pretty complicated.

Mottie:
It would still be totally doable. I think it would be easiest if we had some predefined "font size / line height" settings in a select dropdown (on userstyles).

silverwind:
Font size can break layouts. If we're certain it doesn't break anything, we can go for it.

silverwind:
Though, I'd advice using full page zoom if the font is really unreadable.

PR #234, opened 2015-05-22
Fix pull request icon color

eilgin:
In the notification panel, all PRs are seen in purple, even if they've got different states.
This quick fix resolves this.


silverwind:
Thanks, I think this only applies to unread PRs, which I have none to check right now. But I see the style in GH's sheet, so this should be good.

PR #259, opened 2015-10-24
Commit: fix "committed for" in author section

auscompgeek:
This fixes the "committed for" part of the author section of commits and commit teases, which appears when the committer is not the author.

Example: https://github.com/vim-scripts/taglist.vim

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1508,7 +1508,7 @@
   .dropdown-header, .participation a, .completeness-indicator-info,
   .type-icon-state-none, a.type-icon-state-none, .pagehead-tabs-item,
   a.pagehead-tabs-item, .org-settings-team-count, a.muted-link strong,
-  table.files td.icon, .commit-tease {
+  table.files td.icon, .commit-tease, .commit-author-section {
     color: #808080 !important;
   }
   .vcard-detail .octicon, .member-badge .octicon,


Mottie:
Thanks, that looks good.

silverwind:
I'd go with a bit lighter color. The original is `#333`, the closest inverse color we use would be `#c0c0c0` or `#a0a0a0`.

auscompgeek:
Ah, well spotted. I wasn't comparing it with GitHub's stylesheet. I suppose how bright it should be is arguably a matter of opinion.

Given that GitHub's CSS gives it more contrast than the commit message summary, we should too. I wouldn't go as bright as the usernames though.

auscompgeek:
On this note, it appears that GitHub has decided to, rather than directly style commit timestamps on commit views, indirectly style them through `.commit-author-section` as well.

silverwind:
I usually go with the inverse, maybe a step darker sometimes:

`#000` -> `#e0e0e0`
`#333` -> `#c0c0c0`
`#666` -> `#a0a0a0`
`#888` -> `#808080`

silverwind:
Went with `#a0a0a0` now. `#c0c0c0` would've been the inverse, but it felt a bit too bright.

PR #264, opened 2015-11-17
Update for new GitHub Repository View

mkoch227:
As indicated at https://github.com/blog/2085-a-new-look-for-repositories, a new look for respositories is coming.  Some CSS has changed for the active tab, so this PR updates the CSS to apply the necessary styles for the new CSS.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -592,7 +592,8 @@
   .migrate-org-roles, .migrate-org-roles-header, .migrate-org-roles-item,
   .migrate-org-roles-count, .migration-footer, .migration-feature-list:before,
   .migration-org-avatar, .org-migration-settings-section,
-  .word-upload-callout .note, .commit-tease .commit-meta {
+  .word-upload-callout .note, .commit-tease .commit-meta,
+  .js-selected-navigation-item.selected.reponav-item {
     border-color: #343434 !important;
   }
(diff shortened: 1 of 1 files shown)


github-dark.css
```diff
@@ -658,7 +659,8 @@
   .select-menu-button::before {
     border-top-color: #eee !important;
   }
-  .pagehead-tabs-item.selected {
+  .pagehead-tabs-item.selected,
+  .js-selected-navigation-item.selected.reponav-item {
```
silverwind: Maybe this is overqualified? Does `.reponav-item.selected` suffice?

silverwind: (same applies to all additions in this PR)

silverwind:
Can you make the inactive tabs `#888` instead of the link color? As these are probably `a` tags, you may need to do `a.class` to outweight our `a` style.

mkoch227:
I went ahead and switched the selectors to `.reponav-item.selected` as it also works, and `.js-selected-navigation-item.selected.reponav-item` did seem overqualified. I  also changed the inactive tab's text to `#888` as requested.  I've also included how it looks below in case you're unable to view the new design yet.

screen shot 2015-11-17 at 5 46 53 pm

silverwind:
Thanks :+1:

PR #266, opened 2015-11-18
New repos: Fix repository meta/description

auscompgeek:
The `.repository-description` class no longer exists in the new look, and is also not styled by GitHub in the old look.


silverwind:
Thanks. How did you find out that it's unused? Did you search their CSS?

auscompgeek:
Luckily, I happened to still have a tab open from before I opted into the new look. :)

PR #272, opened 2015-12-13
Darken scroll bars for WebKit-based browsers (Safari, Google Chrome, etc.)

mmizutani:
This tiny snippet darkens both horizontal and vertical scroll bars of github.com on web browsers such as Safari, Google Chrome and Opera except for Firefox.  (Strictly speaking, Google Chrome no longer uses the WebKit engine, but `::-web-kit-scrollbar**` still works on the latest Google Chrome.)

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2089,6 +2089,16 @@
     background: #300 !important;
     border-color: #500 !important;
   }
+  /* scroll bar */
+  ::-webkit-scrollbar {
+      background: #181818 !important;
+  }
+  ::-webkit-scrollbar-track {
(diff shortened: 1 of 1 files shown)


silverwind:
We already have these, but limited to ZenHub it seems:

https://github.com/StylishThemes/GitHub-Dark/blob/12600f65b9defe3febd9f7198b459957637851fd/github-dark.css#L2302-L2314

@Mottie Should we use the ones defined here and remove them from ZenHub?

Mottie:
Sure that works! Thanks @mmizutani!

PR #273, opened 2015-12-14
darken webkit scrollbar corners

mmizutani:
I forgot to include `::-webkit-scrollbar-corner` selector in my last pull request (#272), which has already been committed to the master branch. So I am putting this separate pull request to add it. Without this selector, the bottom right corner (the intersection of the horizontal and vertical scroll bars), which appears only when both scroll bars are active like when Octotree sidebar is opened, does not turn dark gray and remains white. Sorry, I should have tested more thoroughly.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2096,7 +2096,8 @@
     max-width: 12px !important;
     background: #181818 !important;
   }
-  ::-webkit-scrollbar-track {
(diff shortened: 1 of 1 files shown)


silverwind:
@Mottie can you review?

Mottie:
Look good

PR #276, opened 2015-12-21
MediaWiki: Apply background-color to the TOC

SEAPUNK:
Example: https://github.com/bitcoin/bips/blob/master/bip-0037.mediawiki

Before:

After:


silverwind:
Thanks! I'll follow up with adding a border to make it a bit more distinguishable from the background.

PR #278, opened 2015-12-31
Constrict the horizontal scrollbar height

keith:
This makes sure the horizontal scrollbar can not be any taller than 10 pixels

Previously the horizontal scrollbar looked a little large:

screen shot 2015-12-31 at 13 46 14

This makes it a little shorter:

screen shot 2015-12-31 at 13 51 39


Mottie:
Thanks!

PR #300, opened 2016-03-24
Fixed stylelint rules for version 5.

rayrutjes:
According to https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#500

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -6,7 +6,7 @@ rules:
   no-missing-eof-newline: true
   color-hex-length: short
   color-no-invalid-hex: true
-  color-no-named: true
+  color-named: never
   number-leading-zero: never
   number-max-precision: 3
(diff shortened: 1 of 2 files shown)


silverwind:
Thanks for picking up the slack, much appreciated!

@Mottie please don't update to new major versions without reading the changelog please :wink:

Mottie:
I did... umm, but am I missing something?

silverwind:
You updated the dependency in https://github.com/StylishThemes/GitHub-Dark/commit/36adfa9f6e547009f151daf3c52ec2e0b8bf7d46. Major version upgrades ( eg. `4.x.x` -> `5.x.x`) will likely include breaking changes like it was the case here, so it's best to double check everything is working after such an upgrade.

Mottie:
Okie dokie

PR #322, opened 2016-05-31
Fix checklist rendering in issue template

jawshooah:
Without the leading hyphen, these checkboxes are rendered as plaintext.

---
This change is [Reviewable](https://reviewable.io/reviews/stylishthemes/github-dark/322)


Mottie:
Thanks!

Issue #336, opened 2016-06-22
Organization header.


rayrutjes:
![image](https://cloud.githubusercontent.com/assets/5570853/16257044/3893c660-3856-11e6-8153-3ea4225a2b75.png)

Mottie:
Hi @rayrutjes! Please update your style.

rayrutjes:
https://github.com/StylishThemes/GitHub-Dark/issues/335

PR #337, opened 2016-06-27
Fix organizarion header colors

fefas:
Hello,

When I open any organization page, I see the header with a light color:
captura de tela 2016-06-26 as 22 23 58

So, I added the follow lines at the end of my local stylesheet and the problem was fixed:
```
.orghead {
   background-color: transparent;
   border-bottom-color: #333;
}
```
captura de tela 2016-06-26 as 22 24 17

I believe that the end of file is not a good place for my change, but I don't know where it could be.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -3484,3 +3484,8 @@
   }
   /* Literal.Number.Integer.Long */
 }
+
+.orghead {
+    background-color: transparent;
+    border-bottom-color: #333;
+}


fefas:
I see the isse #336 .. I have already updated my style :/

Mottie:
There's something going on with the Chrome Stylish addon. It doesn't seem to auto-updating the styles.

If you look at [lines 961-963](https://github.com/StylishThemes/GitHub-Dark/blob/master/github-dark.css#L961-L963) you'll see that `.orghead` is already defined there with a semi-transparent background.

Please update your style by clicking on the "Update with Stylish" button here: https://userstyles.org/styles/37035/github-dark.

fefas:
I removed and installed the style.. now it works :) .. I am in Chrome and it seems to have a problem when updating .. thank you very much

Mottie:
You could also check out our [userscript](https://github.com/StylishThemes/GitHub-Dark-Script)... it auto-updates at least once a day. And you can use the "Force Update" button to ensure it updates.

PR #344, opened 2016-07-17
User profile tabs: add base-color top border on selected tab

un1versal:
Essentially this, before and after below

tabs


Mottie:
Thanks!

PR #351, opened 2016-07-19
add subdomain: darken https://services.github.com/

un1versal:
like it says on the tin

Seems to work very well

its accessed via Training link on footer

capture


Mottie:
It's not as simple as just including the subdomain... there are a lot of styles on that and related pages that will need tweaking.

un1versal:
Looms pretty good as is with this tbh, but its your call I cant really find anything on there that doesnt work over a full nvigation test on all pages there.

silverwind:
Not sure there's much value in styling that. I think it's highly unlikely that a beginner who does these tutorials will even have our style installed. Also, it's not really something one visits more than once, further reducing value in maintaining a style there.

un1versal:
fine

PR #363, opened 2016-07-25
fuggly town: drawn attention to tag line

un1versal:
capture


Mottie:
I actually feel like it's too distracting now...

*hands Mr I-like-fuggly new hobby*

un1versal:
> hands Mr I-like-fuggly new hobby

Nice self deprecating insults... :D hahaha, priceless

un1versal:
> I actually feel like it's too distracting now...

And I was thinking about adding a animated preview instead of static.... hahaha but tbh I have to be in a special mood for that like suicidal or something due to being bored out of my scull, (which never happens)

PR #358, opened 2016-07-25
not too bright (my corneas cried for help) not too dark

un1versal:
@Mottie this looks good but not so bright that it pierces corneas
in other places slightly too dark on the dull side, check it.

PR #364, opened 2016-07-27
revert part of #358

un1versal:
IMO this whole block of selectors needs segregating selectively, else some
bits are just too dull, however Im not confident enough to handle this task so Ill simply revert the offending portion.

this is in response to  https://github.com/StylishThemes/GitHub-Dark/pull/358#discussion_r72374077


Mottie:
Thanks!

PR #374, opened 2016-08-03
brighten up selected text in navigation

### LINKED ISSUE #373: selected text visibility on label selection menu
### Unstyled Content

- [x] No other existing issue and/or pull request.
- [ ] URL of unstyled content.
- [ ] If unable to provide a URL, please report the class name of unstyled content.
- [x] Provide steps to reproduce (opening a dialog, etc).

### Steps to Duplicate

assign or unassign labels to an issue/pr when hovering over the already set labels you see selected text is too dark.

demo-a

I think class is `.label-select-menu .select-menu-item.selected` https://github.com/StylishThemes/GitHub-Dark/blob/master/github-dark.css#L1681-L1681

un1versal:
closes #373

** so this is my proposal** 

I think this is half decent compromise also brightened the section brightness slightly, could be better I suppose.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1529,8 +1529,9 @@
   a.filter-item.selected, .date-selector td.selected,
   .date-selector td.selectable-day:hover, .subnav-item:hover, .subnav-item:focus,
   .owner-select-target:hover, .owner-select-target:focus,
-  .suggester li.navigation-focus, .suggester li.navigation-focus small {
-    color: #efefef !important;
+  .suggester li.navigation-focus, .suggester li.navigation-focus small,
+  .label-select-menu .select-menu-item.selected, .octicon .octicon-x {
+    color: #f5f5f5 !important;
   }
   .btn, a.btn, .btn.btn-primary, .button.primary, .minibutton.primary,
   .btn-danger:hover, ul.repo-stats li a:hover, .text-emphasized,
@@ -1678,7 +1679,6 @@
   a .mega-octicon, .anchor, a.anchor,
   .gcr-ext-editor-close .delete-button:hover svg,
   .repo-filterer .repo-filter.filter-selected, .select-menu-divider,
-  .label-select-menu .select-menu-item.selected,
   .dropdown-signed-commit .dropdown-menu, .signed-commit-signer-name .signer,
   .signed-commit-signer-name a.signer, .text-gray-dark, a.text-gray-dark,
   .text-gray strong, .display-heading-1, .display-heading-2, .display-heading-3,


github-dark.css
```diff
@@ -1529,8 +1529,9 @@
   a.filter-item.selected, .date-selector td.selected,
   .date-selector td.selectable-day:hover, .subnav-item:hover, .subnav-item:focus,
   .owner-select-target:hover, .owner-select-target:focus,
-  .suggester li.navigation-focus, .suggester li.navigation-focus small {
-    color: #efefef !important;
+  .suggester li.navigation-focus, .suggester li.navigation-focus small,
+  .label-select-menu .select-menu-item.selected, octicon-x {
```
silverwind: Is octicon-x really the right selector? How does Github select/style it?

un1versal: It appears to be (heres a demo with a purple color for added effect)

it also works with ```.octicon .octicon-x``` which is what the actual class said

if its right or wrong... you tell me ;)

demo-a

github-dark.css
```diff
@@ -1529,8 +1529,9 @@
   a.filter-item.selected, .date-selector td.selected,
   .date-selector td.selectable-day:hover, .subnav-item:hover, .subnav-item:focus,
   .owner-select-target:hover, .owner-select-target:focus,
-  .suggester li.navigation-focus, .suggester li.navigation-focus small {
-    color: #efefef !important;
+  .suggester li.navigation-focus, .suggester li.navigation-focus small,
+  .label-select-menu .select-menu-item.selected, .octicon .octicon-x {
```
silverwind: I'd suggest changing it to

```css
.select-menu-item.selected
```

And then add
```css
.select-menu-item .octicon
```
to the group
```css
fill: currentColor;
```

Also, not sure about this color change below. May I suggest just adding it to `#fff` instead?

silverwind: The second part might not even be necessary. If `.octicon` proves to have enough specificity, color will just cascade down to the SVG.

un1versal: > Also, not sure about this color change below. May I suggest just adding it to #fff instead?

yea definitely.

Im not sure what you mean about fill: currentColor; group, I cant see any anywhere.

silverwind: All octicons (which are SVG) on the site have `fill: currentColor`, so you just need to set `color` on any parent element to color them too. If there's no wrongly colored parent in the way, that icon should get it's color inherited from `.select-menu-item.selected` ideally.

Mottie:
Ummm, weird... is anyone else seeing this behavior in Chrome? I disabled all extensions, and it's still happening. It's not happening at all in Firefox.

scroll

silverwind:
Side effect of a Chrome experiment maybe?

PR #390, opened 2016-08-10
fix baby sh*t color styles and make text color over alert boxes consistent

un1versal:
Im sorry, but the color style picked to the red alert boxes is simply awful (there isnt a good compromise that way imo aside from this)

Following GitHub defaults/styles (ON DARK) should only be done with a pinch of salt on very few selected items. (unless you dont care about aesthetics or visibility or users with visual impairments)

this reverts the last two commits and add the missing selector `.comment-form-stale` to same style group as `comment-form-error`

@silverwind is bound to disagree with this, because he loves this awful stuff and swears by it, but its simply unreadable and his attempts to make it readable made it look simply awful.

and appeal to common sense @Mottie


Mottie:
I'm just going to close this...

un1versal:
fuckitall

Take 5 of those 3 times a day and maybe you come close to feeling what I feel about that decision.

🗡

Issue #388, opened 2016-08-10
Impossible to read header colors on elements and pages


un1versal:
### Unstyled Content

- [x] No other existing issue and/or pull request.
- [ ] URL of unstyled content.
- [ ] If unable to provide a URL, please report the class name of unstyled content.
- [x] Provide steps to reproduce (opening a dialog, etc).

Affected pages (random content inside)

https://github.com/site/terms
https://github.com/site/privacy
https://github.com/security

https://github.com/explore
https://github.com/open-source
https://github.com/business

the gif below is incomplete thers way too much stuff to include

![demo-a](https://cloud.githubusercontent.com/assets/3521959/17546216/faf2541c-5ed8-11e6-8ddf-236489af9b06.gif)

The one of many examples to show @silverwind that GitHub styles are flawed especially when you just darken something and leave content untouched or pick color styles that are equally as terrible.

i.e. You cannot follow these styles blindly. **CONSIDER USERS WITH VISUAL IMPAIRMENTS** (pun intended.) and ffs dont use baby sh*it color styles it looks awful on DARK

![capture](https://cloud.githubusercontent.com/assets/3521959/17546552/ae743edc-5eda-11e6-9108-2ac8b5f56f1c.PNG)

![capture3](https://cloud.githubusercontent.com/assets/3521959/17546681/450bbb5e-5edb-11e6-9380-6980d11ce309.PNG)

![capture2](https://cloud.githubusercontent.com/assets/3521959/17546680/4505baec-5edb-11e6-86c1-9c15ca51123d.PNG)

Mottie:
I think I've fixed most of these... I don't remember how to get to this page though:

![capture](https://cloud.githubusercontent.com/assets/3521959/17546552/ae743edc-5eda-11e6-9108-2ac8b5f56f1c.PNG)

un1versal:
> I think I've fixed most of these... I don't remember how to get to this page though:

https://github.com/contact

Mottie:
I'll leave this issue open for now in case you find any more header color issues.

silverwind:
I think we should eventually create separate `@-moz-document` sections for sites that run off their own stylesheets like `help.github.com`. Does Stylish for Chrome support multiple `@-moz-document` rules when importing a Mozilla-format style yet?

Mottie:
Yes, Stylish splits them into separate sections. So it works just fine. Using the built-in editor is a tiny bit more troublesome though.

[...]

Issue #333, opened 2016-06-21
Unstyled content on dropdown menu


J2TeaM:
### Unstyled Content

- URL of unstyled content: https://help.github.com/
- CSS selector: `.dropdown-menu:before`
- Screenshot: 

![image](https://cloud.githubusercontent.com/assets/5250117/16239853/038ae77c-3811-11e6-9f8b-53be2884ad5d.png)

Mottie:
Thanks for reporting this problem! It's been fixed in the repo, but not yet on userstyles.org.

PR #393, opened 2016-08-17
Set Org Name Text Color

geekgonecrazy:
Sets it to look like:
image

Instead of:

image


Mottie:
Hi @geekgonecrazy!

Thanks for the contribution, but I think it might be better to add back some of our header selector color definitions (e.g. `h1, h2,...`) - what do you think @silverwind?

xt0rted:
I'm running a fork of this in Edge and had to add in the following for the org page header:
```css
h1.org-name {
    color: #fff !important;
}
```

When I ran into this I thought it was working in FireFox but looking just now shows it's not. Maybe it was changed recently.

geekgonecrazy:
Possibly :smile: That's the only one I had seen that was really bugging me in my daily workflow.

I saw some discussion going on in #388

silverwind:
> I think it might be better to add back some of our header selector color definitions (e.g. h1, h2,...) - what do you think @silverwind?

I wouldn't. It's probably going to cause more issues in the long run to have different selectors than GitHub.

@geekgonecrazy care to move `.org-name` to `#c0c0c0`? It's our counterpart to GitHub's `#333`.

geekgonecrazy:
@silverwind done.

Mottie:
Thanks!

PR #397, opened 2016-09-03
signup button: consistent styling normal + hover is green.

un1versal:
Unstyled
capture

Current styled
capture-current

Fixed
capture-after


Mottie:
Thanks!

PR #222, opened 2015-04-02
Changed the 'private' tags to black

EvanDarwin:

PR #400, opened 2016-09-06
fix background in payments for failed/refunded txn

yaauie:
Fixes the table row background for `.failed` and `.refunded` rows within
the `.payment-history` table available on https://github.com/settings/billing

before:
screen shot 2016-09-06 at 22 09 21z

after:
screen shot 2016-09-06 at 22 09 01z

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -942,7 +942,8 @@
   .ajax-pagination-form .ajax-pagination-btn:hover,
   .ajax-pagination-form .ajax-pagination-btn:focus,
   .manage-repo-access-not-active, .header-search-scope, .reverse-progress-bar,
-  .featurette.py-5, .ace_button:hover {
+  .featurette.py-5, .ace_button:hover,
+  .payment-history .refunded, .payment-history .failed, {
     background: #222 !important;
   }
   hr, .rule, #browser, .repos, .options-group, .pagehead ul.tabs li a.selected,


Mottie:
Hi @yaauie!

I think `#333` might be a bit too bright for those rows. Try using `#222` and please add the selectors to the *bottom* of the list of selectors. We try to keep the newest selectors closest to the definition.

Thanks!

yaauie:
@Mottie I've updated my patch; `#222` does look a lot better 👍 
screen shot 2016-09-06 at 23 46 47z

Mottie:
Awesome! Thanks!

github-dark.css
```diff
@@ -942,7 +942,8 @@
   .ajax-pagination-form .ajax-pagination-btn:hover,
   .ajax-pagination-form .ajax-pagination-btn:focus,
   .manage-repo-access-not-active, .header-search-scope, .reverse-progress-bar,
-  .featurette.py-5, .ace_button:hover {
+  .featurette.py-5, .ace_button:hover,
+  .payment-history .refunded, .payment-history .failed, {
```
Mottie: For future reference, please don't include trailing commas. Luckily I remembered to use the linter and I fixed it before releasing a new update 😉.

silverwind: I'm surprised it caught that. Do you recall which rule had been hit? The rule which is supposed to catch this is yet to be written (https://github.com/stylelint/stylelint/issues/1777).

Mottie: I tested it again on the cusom code font line and found out that the error is because I ran perfectionist before lint.

Before perfectionist (no lint error)

```css
, .blob-code-inner, {
```

After Perfectionist

```css
, .blob-code-inner,  {
```

perfectionist adds an extra space, so the linter threw this error:

> github-dark.css
 19:34  ×  Expected single space before "{"   block-opening-brace-space-before

>\>\> Exited with code: 1.
Warning: Task "exec:stylelint" failed. Use --force to continue.

> Aborted due to warnings.

silverwind: How nice of Perfectionist 😆 Quite an unexpected interaction.

PR #406, opened 2016-09-15
Project: Fix triage panel color

40n:
index


Mottie:
Hi @40n!

Thanks for your contribution, but I think that panel would look better with a solid background.

PR #421, opened 2016-09-25
Ace editor: Fix a white border

40n:
index

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -882,6 +882,13 @@
     background: #181818 !important;
     border-color: #484848 !important;
   }
+  .CodeMirror-gutters {
+    background: #1d1f21 !important;
+    border-color: #484848 !important;
+  }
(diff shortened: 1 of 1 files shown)


silverwind:
That's not ACE, it's CodeMirror. I guess this means we have to redo all the ACE themes now. I wonder what prompted them to switch. There's a `cm-s-github` theme active which seems to be rather minimal, e.g. most of the styles come from the default CodeMirror theme.

I wonder if we can drop ACE themes. I checked Gist and GitHub which both have been switched to CodeMirror.

silverwind:
Regarding this PR: Can you use the existing `181818` selector groups (one with border, one without) instead of introducing a new color?

silverwind:
On the other hand, I think the changes surrounding CodeMirror require a bit more planning so we can support multiple themes. I'll close this now and will follow up in https://github.com/StylishThemes/GitHub-Dark/issues/422.

PR #426, opened 2016-10-01
Added Monokai - Spacegray Eighties theme from Sublime Text

torrentails:
tmTheme source:
https://github.com/pyoio/monokai-spacegray

It seems GitHub doesn't apply separate css classes some elements that it should and so the theme, unfortunately, isn't exactly like it appears in Sublime but it's close enough.

Only tested the appearance with Python, javascript and xml files but I think it should look fine with other languages on GitHub.


Mottie:
Wow, thanks for all the awesome work! We really appreciate it!

PR #430, opened 2016-10-13
move comments  previous to occurrence they belong to

un1versal:
this makes more sense to me.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2798,230 +2798,232 @@
     background: #141414 !important;
     color: #ccc !important;
   }
+   /* comment */
   .blob-code, .blob-code-inner {
     color: #ccc !important;
   }
+  /* constant */
   .pl-c, .pl-c span {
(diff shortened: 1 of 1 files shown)


silverwind:
Note that the CSS there is just copy-pasted from the theme source files [like this one](https://github.com/StylishThemes/GitHub-Dark/blob/master/themes/github/twilight.css), where it does make more sense (comments are on the same line). What you're seeing is `perfectionist` formatting doing a bad job on these comments.

I'd say we're better of just stripping these comments entirely before pasting theme into the file. For development, one can look up the classes in the individual theme source files.

un1versal:
Ill see if I have patience to strip them all together instead and force push changes, wont be today though.

silverwind:
I'm not sure this is worth it. I'd be fine if you just copy-paste the all-in-one-line style from the Twilight theme into `github-dark.css`.

PR #432, opened 2016-10-13
fix contrib calendar heats and cleanup

un1versal:
First commit fixes this

store

second commit rearranges selectors

3rd commit fixes blame heats

store

Of course this could be fixed as it was before using base color but that is for whoever wants to figure that bit out and in that case close this.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1132,7 +1132,7 @@
   .calendar-graph rect[fill="#d6e685"],
   .calendar-graph rect[fill="#8cc665"],
   .calendar-graph rect[fill="#44a340"],
-  .calendar-graph rect[fill="#1e6823"], .heat {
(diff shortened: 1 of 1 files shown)


silverwind:
Not sure I follow on how your fix is supposed to work. When I change the base color manually, both color legends look alright:




Wouldn't removing the selectors result in the default site colors being displayed?

un1versal:
the gifs are pretty clear of before and after, atm as is its mixed for me.

silverwind:
The fallback color was off to our default `#4183c4` which I've fixed. Maybe that was the reason for your issue?

un1versal:
that fixes nothing sadly.

un1versal:
this issue is back with a vengeance, it now all yellow.

PR #433, opened 2016-10-14
rearrage calendar contrib heat section

un1versal:
simple cleanup to help readability

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1129,9 +1129,11 @@
     border-color: #484848 !important;
   }
   /* contribution calendar and blame heats */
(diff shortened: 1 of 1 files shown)


silverwind:
Our reformatter `perfectionist` might still put it in multiple lines when we decide to run it, but for now it'll stay.

un1versal:
Sounds to me perfectionist is not

silverwind:
Yes, not a big fan of it because it isn't configurable enough. I'd gladly drop it if there were some better alternative.

Oh, and I'll repubish again, seems the site it messed up once again during publish.

silverwind:
Published, looks good I think.

PR #436, opened 2016-10-19
Fix header color

40n:
index

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -718,7 +718,7 @@
   .discussion-item + .discussion-item,
   .discussion-item-review + .discussion-item, .libraries-jumbotron,
   .last-review-thread, .menu-item, .review-comment.is-comment-editing, .Box,
-  .Box-footer, .team-list-footer {
+  .Box-footer, .team-list-footer, .Box-header {
     border-color: #343434 !important;
   }
   .discussion-item-icon, .date:after, .render-shell img, img.asset,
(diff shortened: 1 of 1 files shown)


github-dark.css
```diff
@@ -646,7 +646,7 @@
   }
   pre, h1, h2, h3, .header, table, table tr, table td, table th, blockquote,
   .pulse-graph, .btn:not(img), .minibutton, .social-count,
-  .discusion-topic-infobar, .box-header, .box-body, .timeline-comment-label,
+  .discusion-topic-infobar, .Box-header, .box-body, .timeline-comment-label,
```
silverwind: You can remove this selector, it gets outruled by the rule you added below.

40n: https://github.com/40n/GitHub-Dark/commit/1f907f2bb3259a2161dd91d39ab144a131304710

silverwind: You can push to the same branch `fix`, changes will update in GitHub.

github-dark.css
```diff
@@ -641,7 +641,7 @@
   .gist-quicksearch-result, .gist-quicksearch-result-group, .listgroup,
   .listgroup-item + .listgroup-item, .user-key-badge, .newsfeed-placeholder,
   .progress-bar-inline .progress-bar, .menu-heading, .border-gray-dark,
-  .discussion-item-review .blob-wrapper {
+  .discussion-item-review .blob-wrapper, .Box-header {
```
silverwind: You can remove this rule. The rule below has the same specificity (1 class, 1 !important) so the last rule will always win.

40n: Sorry, I was only half awake when I wrote this.

Issue #395, opened 2016-08-26
Network Graph not displaying at all


retep998:
### Unstyled Content

- [x] No other existing issue and/or pull request.
- [x] URL of unstyled content. https://github.com/StylishThemes/GitHub-Dark/network or similar on any other repo.
- [x] Provide steps to reproduce (opening a dialog, etc). Go there with Firefox 47 on Windows 10 and the Network Graph doesn't display. Disabling the style with Stylish makes the Network graph visible again, reenabling the style promptly makes it vanish again.
- [x] Screenshots.
 Github Dark enabled ![](http://i.imgur.com/rdRdMek.png)
 Github Dark disabled ![](http://i.imgur.com/yooL8Ca.png)

Mottie:
Hmm, I can't seem to duplicate this... I'm using Firefox v48.0.2 in Windows 10.

Are you using Stylish or the script?

silverwind:
Also cannon reproduce (on Firefox 50 right now). Can you right-click > inspect the area and see if there's anything there at all, and if some style is hiding it?

retep998:
I am using Stylish.

After a quick test with setting background-color on stuff...
The `<div id="network" class="network">` element is definitely visible.
The `<div class="network-graph-container js-network-graph-container" data-network-graph-meta-url="/StylishThemes/GitHub-Dark/network/meta" data-network-graph-chunk-url="/StylishThemes/GitHub-Dark/network/chunk">` inside it definitely is not.

Like, the canvas is there, but I have no idea why it isn't rendering. My cursor changes to the 4 way move cursor when I hover over the area where it should be, and turning off ad blockers has no effect. I don't really know what is going on. It is clearly there, just invisible. I can even click on it and it'll take me places.

Mottie:
Try disabling all of the other addons. If it reappers then you know the culprit isn't Stylish. Re-enable the plugins one-by-one to find the cause.

If the problem is within Stylish, the try disabling the other styles.

retep998:
I have every single style disabled except for Github Dark (not that I had any other styles that applied to Github anyway). I have disabled all my addons, switched back to default skin for Firefox, and restarted my browser. Network Graph still does not display with Github Dark enabled via Stylish. Disabling Github Dark makes it appear, and enabling it makes it disappear.

retep998:
Here is the full contents of my current Github Dark style: https://gist.github.com/retep998/036bb8377e6648073c79ca3f33d51c07

[...]

PR #446, opened 2016-11-26
Dark theme for network graph

j0yu:
image

Much but much better than the current soft grey

image

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1401,7 +1401,6 @@
     -webkit-filter: invert(90%) hue-rotate(180deg) saturate(300%) !important;
             filter: invert(90%) hue-rotate(180deg) saturate(300%) !important;
   }
-  }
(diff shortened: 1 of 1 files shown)


Mottie:
Hi @j0yu!

We used to use `hue-rotate` to get some nice colors, but it was causing some troubles in Firefox & GPU rendering - see #395. Since Firefox v50 is out, maybe this problem has been resolved? Pinging @retep998 to see if applying this patch still causes the same issue.

j0yu:
@Mottie Thanks for the feedback, sorry I didn't spot that. 

Maybe this isn't a good idea then for public release after reading #395, I'll keep this feature for private use for the time being

(Closing pull request)

Mottie:
You could add a custom Stylish style with just these definitions. That way you won't have to modify GitHub-Dark when it updates.

j0yu:
Great idea. Just done it, thanks for the suggestions!

PR #468, opened 2017-03-05
Star color in repo list is gray instead of bluish.

tyler-sommer:
In the repo list on your user dashboard, the star counts and icon is a blue-ish tinted gray. The repo icon is `#777` and the proximity of the two barely different grays looks a bit strange.

This PR changes the star counts and colors to `#777` which is the same as the repo icons.

Here is a before pic:
screen shot 2017-03-05 at 12 51 07 pm

And after:
screen shot 2017-03-05 at 12 52 48 pm

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -500,6 +500,10 @@
   .discussion-topic-header, .previewable-comment-form .tabnav {
     border-radius: 3px 3px 0 0 !important;
   }
+  /* star count and color in repo lists */
+  .mini-repo-list-item .stars {
+	  color: #777;
+  }
   /* repo labels, private repo lists */
   .repo-label span, .private .repo-list-item, .hook-delivery-guid {
     background: #222 !important;


github-dark.css
```diff
@@ -500,6 +500,10 @@
   .discussion-topic-header, .previewable-comment-form .tabnav {
     border-radius: 3px 3px 0 0 !important;
   }
+  /* star count and color in repo lists */
+  .mini-repo-list-item .stars {
+	  color: #777;
+  }
```
silverwind: Can you add your selector to [this line](https://github.com/StylishThemes/GitHub-Dark/blob/88e24df93005d56afde31d03aef6f172dfc799ed/github-dark.css#L1933) instead? We try to conserve space by grouping similar colors together and I think `#7b7b7b` is reasonably close to your `#7777`.

silverwind:
Thanks. I too noticed these near-gray blue shades lately on various pages. We should probably map them all to `#7b7b7b`.

silverwind:
Fixed the issue with near-gray colors through https://github.com/StylishThemes/GitHub-Dark/commit/8a3936a6d348ad5c4ee082419e7097f036d88ca7 and https://github.com/StylishThemes/GitHub-Dark/commit/383826a3eaa0aa2d21efbcd5d208158859d461d2.

PR #473, opened 2017-03-22
Fix zenhub burndown page

stedesign:
I've tried to fix the Burndown page for Zenhub, below the result

fix


Mottie:
Hi @stedesign!

Thanks for your contribution!

Notes on this PR:

* As much as possible, please combine definitions with close-to or similar colors.
* All newly added definitions should be appended to the list of selectors (next to the open curly bracket) so we can get a general idea of which selectors may have changed and may be removed.
* Lastly, there is no need to add yourself to the authors file as the file is automatically generated 😸

github-dark.css
```diff
@@ -2371,8 +2371,8 @@
   .render-shell img:not(.math) {
     background-image: url("data:image/gif;base64,R0lGODlhCgAKAIAAABoaGgAAACwAAAAACgAKAAACEYQdmYcaDNxjEspKndVZbc8UADs=") !important;
   }
-  .render-shell:hover .img-view img, .file:hover .image img, .file:hover .image canvas, img.asset:hover,
-  .render-shell:hover img:not(.math) {
+  .render-shell:hover .img-view img, .file:hover .image img,
+  .file:hover .image canvas, img.asset:hover, .render-shell:hover img:not(.math) {
```
Mottie: Please add all new selectors next to the `{`, instead of before other (older) selectors.

github-dark.css
```diff
@@ -2733,6 +2737,19 @@
...
+  .zhc-progress-bar > .zhc-progress-bar--completed-issues {
+    background: #163 !important;
+  }
+  .zhc-item-table__header {
+    color: #e2e2e2 !important;
+    background: #202020 !important;
+  }
+  .zhc-item-table__cell__issue-number {
+    color: #7b7b7b !important;
+  }
```
Mottie: While we are keeping the ZenHub definitions separate from the main page definitions, please try to combine these new additions with other ZenHub definitions... with same, or nearly-similar colors. Thanks!

github-dark.css
```diff
@@ -2743,7 +2760,8 @@
   .zh-epic-creator-nav, .zh-epic-creator-picker-container, .zh-board-menu,
   .zh-epic-creator-filters, .zh-epic-issues-picker-new-issue, .merge-status-list,
   .zhc-dropdown-simple-item, .zhc-pipeline-body, .zhc-board-settings-head,
-  .zhc-pipeline-sort-status {
+  .zhc-pipeline-sort-status, .zhc-chart-container, .zhc-burndown-chart__stats,
+  .zhc-chart-container__footer {
     border-color: #343434 !important;
```
Mottie: Perfect! 😉

github-dark.css
stedesign: I haven't touched those classes... I think `grunt clean` made this change

Mottie: Oh, you're right... it shifted the selectors down. Sorry.

Mottie:
Thanks again!

PR #494, opened 2017-05-21
Fix Network Graph

friendlyanon:
Network Graph was left HOT BURNING WHITE AH MY EYES

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1235,6 +1235,11 @@
     -webkit-filter: invert(100%) hue-rotate(180deg) !important;
             filter: invert(100%) hue-rotate(180deg) !important;
   }
+  /* invert network graph */
+  div#network.network > div > canvas {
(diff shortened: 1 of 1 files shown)


silverwind:
Wow, this looks pretty nice. I've attempted to fix this graph multiple times but font rendering was affected by the inverts, so I didn't go for it. Doesn't seem to be an issue with your approach:

friendlyanon:
It's not a perfect solution. Your greyscale avatar for example isn't affected by `hue-rotate`, so it'll stay inverted, but it's still better than what the graph looked like before.
For a perfect solution one'd have to tap into the Javascript core of GitHub, but that'd only be feasible with hijacking `prototype`s and stuff because it's all self contained in anonymous functions.
¯\_(ツ)_/¯

silverwind:
I'm still hoping that they rewrite this to not use `<canvas>` sometimes, so it can be properly be targeted through CSS.

Issue #444, opened 2016-11-16
Unstyled Merge Status Items


Rockstar04:
### Unstyled Content

- [X] I am using the most up-to-date GitHub Dark style!
- [X] No other existing issue and/or pull request.
- [X] Provide the URL of the unstyled content.
        https://github.com/dev-sec/chef-nginx-hardening/pull/39#issuecomment-260933420

![screenshot_2016-11-16_08-08-11](https://cloud.githubusercontent.com/assets/641721/20350419/74ace356-abd4-11e6-8f1f-5a403cfb2e1f.png)

Snippet of HTML code is in a Gist:
https://gist.github.com/Rockstar04/afdadbb735ec8d8c90bfe424f7280306

The specific class that needs attention seems to be  `.merge-status-item`

silverwind:
I'm not seeing it. Is your style up to date? Try updating on userstyles.org.

<img width="934" alt="" src="https://cloud.githubusercontent.com/assets/115237/20353983/efc2eb42-ac1c-11e6-846b-bb4d4bdb00d7.png">

Rockstar04:
I updated when I saw the issue, do you need to restart Chrome for the updated style to take effect? I obviously refresh the page to check if it was corrected.

silverwind:
Should work without a restart. Try editing the style to see the version in the comment. You can also try reinstalling.

Rockstar04:
Sorry, I updated/check for updates several times before reporting.Per your suggestion above, I copied the style manually over from github to the edit screen in Stylish, and then updated again and finally got the latest version. Everything is rendering as expected now.

Mottie:
No worries, it's actually a Stylish for Chrome issue (I'm guessing you're using Chrome):

See https://github.com/stylish-userstyles/stylish-chrome/issues/179.

You can also check out our [GitHub Dark (user)Script](https://github.com/StylishThemes/GitHub-Dark-Script) which allows you to always get the latest update.

Rockstar04:
I read that when I was filling out the template for the issue, and like I said, I clicked check up updated manually, it found an update and I installed it. I did that at least twice before reporting the issue.

Thanks to everyone who works on this! This is the only reason I use Stylish. Github should offer this as a native option.

silverwind:
I assume it's a Stylish bug, but the chances getting that fixed are pretty low as the Stylish author has pretty much lost interest in further development. While the Firefox version is pretty flawless, its days are numbered too.

[...]

PR #507, opened 2017-06-19
Zenhub improvements, mostly for dependencies

olets:
### 1. Styles the Dependency function:

Before:

screen shot 2017-06-18 at 23 23 07

screen shot 2017-06-18 at 23 20 16

screen shot 2017-06-18 at 23 20 00

After

screen shot 2017-06-18 at 23 23 28

screen shot 2017-06-18 at 23 25 03

screen shot 2017-06-18 at 23 13 39

### 2. Also styles the to-do dropdown menu button

Before:

screen shot 2017-06-18 at 23 18 34

After:

screen shot 2017-06-18 at 21 49 47

### Notes
- The colored text in the "this issue is blocked by" doesn't actually change between the before and after - that's just me switching from a red theme to a blue theme
- Something strange is happening with https://user-images.githubusercontent.com/3282350/27267371-e0aace98-5475-11e7-8b2f-dc774eaf5fd3.png (). It should turn that icon in the depency-selector item white when the item is active, but for me in situ it's rendering as black. The documentation asks for attachment links, but it might be best to encode it
- it's possible `.zhc-selection-item--is-active` in line 2704 is made unnecessary by the new necessary `.zhc-selection-item.zhc-selection-item--is-active:not(.zhc-selection-item--is-disabled)` in 2705

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2702,6 +2702,7 @@
   }
   /*** ZenHub styling ***/
   .zhc-selection-item:hover, .zhc-selection-item--is-active,
+  .zhc-selection-item.zhc-selection-item--is-active:not(.zhc-selection-item--is-disabled),
   .zh-select-menu-item.navigation-focus, .zh-select-menu-item:hover,
   .zh-tabs__nav-item:hover {
     background: /*[[base-color]]*/ #4183c4 !important;
@@ -2710,6 +2711,10 @@
   }
   .zhc-pipeline-sort-status__clear {
     color: /*[[base-color]]*/ #4183c4 !important;
+  }
+  .zhc-dependency-type-dropdown--is-open .zhc-dependency-type-dropdown__control {
+    border-bottom-left-radius: 0 !important;
+    border-bottom-right-radius: 0 !important;
   }
   /* background */
   .zhc-pipeline, .zh-epic-creator-top, .zh-epic-creator-nav, input.zhc-icon,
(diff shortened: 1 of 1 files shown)


Mottie:
Hi @olets!

This PR looks good. Thanks so much! I haven't been using ZenHub too much lately, so I haven't kept the style up-to-date.

> icon in the depency-selector item white when the item is active

We have a selector `.zhc-icon:not([class$="-red"]):not([class$="-white"])` that inverts that icon because it has a class that uses "-light". I think adding an extra `:not()` would be necessary, but it'll likely cause problems with their "zh-octicon-light-bulb" icon wherever that might be.

> it's possible .zhc-selection-item--is-active in line 2704 is made unnecessary

It is difficult to test every scenario, it's easier to leave it in there for now.

There are a few more tweaks that need to be made, but I'll work on them after I merge this PR, and when I get some more free time later today. Thanks again!

olets:
This "dependency" function is just a few weeks old, so GitHub-Dark isn't too far behind.

One thing I wondered about is the background color of the "this issue is blocked by" banner.  ZenHub's using a new beige background that I don't think they've used elsewhere. I went ahead and used GitHub-Dark's standard background, but to respect ZenHub's art direction it should be brighter. Is there a GitHub-Dark color that would be fitting? Fwiw this is what the new "blocked" label looks like

screen shot 2017-06-19 at 08 28 56

Re: `.zhc-selection-item--is-active`: that was my thinking too - just wanted it to be documented somewhere.

Re: the icon: Aha! Didn't pick up on that. Probably won't get a chance to play with this today, but the `:not()` sounds like the solution. (Or I suppose we could use a black image)

Thanks for a really great extension!

Mottie:
> but to respect ZenHub's art direction it should be brighter. Is there a GitHub-Dark color that would be fitting

We include several background colors between `#222` and `#444`, so take your pick.

PR #509, opened 2017-07-05
Support GitHub Enterprise highlight

carymrobbins:
I'm not sure if it's due to the version of GitHub Enterprise I'm using, but code line links don't use the appropriate background-color due to a mismatch in the element-level style being applied (too bad github isn't doing this with a css class!).  This patch resolves that issue.


Mottie:
Hi @carymrobbins!

Thanks for the contribution!

PR #512, opened 2017-07-13
Fix highlighted

### LINKED ISSUE #511: Unreadable highlights
Noticed that highlights recently got unreadable without hovering over the text. Using the Monokai theme but is seems to be the same with a few other themes.

Using the Github-Dark script and tried Force Update Style

* **Browser**: Chrome 59.0.3071.115

* **Operating System**: Windows 10 1607
* **Screenshot**:

image

````html
<tr>
        <td id="L95" class="blob-num js-line-number" data-line-number="95"></td>
        <td id="LC95" class="blob-code blob-code-inner js-file-line highlighted">            <span class="pl-k">if</span> (scmSource<span class="pl-k">.</span>buildBitbucketClient() <span class="pl-k">instanceof</span> <span class="pl-smi">BitbucketCloudApiClient</span>) {</td>
</tr>
````

casz:
Fixes #511 

Before:
image

After:
image

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2560,7 +2560,7 @@
   .highlight td.blob-code-inner[style^="background-color: rgb(255, 251, 221)"],
   .highlight td.blob-code-inner[style^="background-color: rgb(248, 238, 199)"],
   .line[style^="background-color: rgb(255, 255, 204)"],
-  td.selected-line.blob-num, td.selected-line.blob-code {
+  td.selected-line.blob-num, td.selected-line.blob-code, td.blob-code-inner.highlighted {
     background: rgba(85, 85, 85, .4) !important;
     border-color: rgba(85, 85, 85, .4) !important;
   }


casz:
Can confirm it fixes it by using `root = "https://raw.githubusercontent.com/casz/GitHub-Dark/patch-1/"` in the user script 👍

auscompgeek:
I wonder if the selectors targeting the style attributes can be removed now.

casz:
I would say definite
image

auscompgeek:
I wouldn't call that definitive. There are a lot of places on GitHub where you can highlight lines.

casz:
Or actually no, because this also affects the diff selected line

Mottie:
Hi @casz!

Thanks!

casz:
Your welcome 🚀

silverwind:
> I wonder if the selectors targeting the style attributes can be removed now.

I'd keep it for a while. Past experience has shown that sites like Gist and Enterprise maintain a separate set of CSS that may still need those old styles. Especially Enterprise is problematic as we can't really test it without a subscription.

PR #518, opened 2017-08-10
Don't list lockfiles in .gitignore, remove them before commits

luftywiranda13:
**What**:
Integrate [remove-lockfiles](https://github.com/luftywiranda13/remove-lockfiles)

**Why**:
Currently this project ignores lockfiles (package-lock.json & yarn.lock). **This can lead to harder debugging process** because contributors might be locked to different versions of deps

**How**:
- Install `remove-lockfiles` then integrate it in `precommit`
- Remove lockfiles from `.gitignore`

**How did I test this?**
- There are tests in `remove-lockfiles` repo
- Also tested in this project by installing deps using `yarn` and `npm`. Lockfiles were generated but not included in commits.
- `remove-lockfiles` not only unstages but also removes the lockfiles before commits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,10 +4,6 @@ build.json
 *.build.min.css
 themes/build/
 
-# lockfiles
-package-lock.json
-yarn.lock
-
 # debug files from npm/yarn
 npm-debug.log
 yarn-error.log
--- a/package.json
+++ b/package.json
@@ -17,9 +17,11 @@
     "grunt-exec": "^3.0.0",
     "grunt-string-replace": "^1.3.1",
     "grunt-wrap": "^0.3.1",
+    "husky": "^0.14.3",
     "parse5": "^3.0.2",
     "perfectionist": "^2.4.0",
     "postcss": "^6.0.8",
+    "remove-lockfiles": "^0.1.0",
     "semver": "^5.4.1",
     "stylelint": "^8.0.0",
     "stylelint-config-standard": "^17.0.0"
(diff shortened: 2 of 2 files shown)


silverwind:
Hmm, I'd rather avoid having them created in the first place than to introduce a git hook. I have to investigate if adding a `.npmrc` and a `.yarnrc` in the project directory can effectively prevent their creation.

luftywiranda13:
@silverwind yes, it can. but the downside is it means you limit the contributors to use a specific package manager only (npm or yarn).

**how about having both of the dotfiles (.yarnrc and .npmrc) together?**
hmmm I don't think it's a good idea to do that. two additional dotfiles just to prevent the lockfiles 😢

silverwind:
Yes, it's a shame. I'd really like a option in `package.json` to disable lockfiles which all package managers should honor.

luftywiranda13:
I agree, wish there's a universal option that we can add in `package.json` to prevent lockfiles creation. But I think we're still far from that because both (npm and yarn) suggest to commit the lockfiles for both libraries and apps

silverwind:
I opted for the [config file](https://github.com/StylishThemes/GitHub-Dark/commit/4f1ffcb5b3f289166db4045894fff1fbece783d0) approach, thought I don't really think the issue will matter. We don't get many contributors and most of them probably never install dependencies.

luftywiranda13:
well yeah it's okay, thanks anyway 😄

PR #567, opened 2017-11-30
Added One Dark Theme from Atom

sparcut:
Source: https://github.com/atom/one-dark-syntax

Translated by hand. 😅 Phew!

Lemme know if any semantic changes needed.

Couldn't find the task to built combined min file for a theme, and I'll just leave the author file and version bumping for you. 

As far as I can tell everything is correct (as possible). From past PR's it seems it's difficult to get 100% perfect translation with compatible GH/CM/JYPT. I've tried to make it as consistent as possible. 

I've looked through things and compiled a list below of a bunch of random languages. For the most part, most work `GitHub` <--> `CodeMirror` with a few exceptions. However, I'd be surprised if anyone would mind, let alone rely on editing projects in the browser, so for looking over and general viewing its 👌.

- JavaScript ✔️
- JSON ✔️
- Python ✔️
- Ruby ✔️
- CSS ✔️
- LESS ✔️
- SASS 🔸
- Stylus ✔️ (Code mirror doesn't support though)
- XML/HTML ✔️
- Java ✔️
- C 🔺 (Substantial differences)
- C 🔺 (Substantial differences)
- C# 🔺 (Substantial differences)
- OCaml 🔸
- Go ✔️
- Perl ✔️
- Rust ✔️
- PHP ✔️
- Shell Script ✔️

I might look into adding the Jupyter version, but for now, I'll leave it at this while I look into Jupyter itself. 👍


silverwind:
> Couldn't find the task to built combined min file for a theme

I think `grunt themes` should handle that.

Mottie:
Hi @sparcut!

Thanks for your hard work! I know it's a pain to tweak and adjust colors for the themes.

I'll take a look at this PR later tonight when I get a chance to look at the updated demo.

Mottie:
Thanks again, it all looks good!

Issue #522, opened 2017-08-25
Theme Request: Material Theme


unigazer:
Edited to show progress:

* [ ] GitHub: **needed** (see #568)
* [x] CodeMirror: Added
* [ ] Jupyter: **needed**

----

Hey, I know I've already asked for one theme from you, `Monokai` and you implemented that the very next day (Thank you on that!).

I was thinking, if there is a possibility, if you could implement this theme for syntax theme:
https://github.com/equinusocio/vsc-material-theme
or something like `One Dark` syntax theme from `Atom`, to at least have `Google Material` colors.

I'm sorry if I'm annoying, it's just that I really like your theme. I mean, my eyes love it, but so do I, too! 😁

Mottie:
Hi @VladimirDev93!

We're all super busy, so if you or someone else finds the time to adapt the material theme to GitHub's PrettyLights highlighter, that'd be awesome!

* [ ] GitHub: **needed**
* [ ] CodeMirror: [Available](https://codemirror.net/demo/theme.html#material).
* [ ] Jupyter: Can be adapted from here: [mkdocs-material](https://github.com/squidfunk/mkdocs-material/blob/master/src/assets/stylesheets/extensions/_codehilite.scss).

I can start working on including the CM and Jupyter themes.

VladimirDev93:
Um... I would like to help, it's just I don't know how to help. Any links on what should I do in order to help?

Mottie:
Well, it's a bit complicated... we should have set up a testing platform to make editing & tweaking stylesheets a bit easier.

You can grab one of the template files from the [`src` folder](https://github.com/StylishThemes/GitHub-Dark/tree/master/themes/src) and tweak it to match the theme colors.

Right now, you can test the tweaks for Jupyter & CodeMirror and GitHub diff pages by downloading this repository as a [zip file](https://github.com/StylishThemes/GitHub-Dark/archive/master.zip) and adding the tweaked template(s) to the [index.html](https://github.com/StylishThemes/GitHub-Dark/blob/master/index.html) file to load it and see what it looks like; but because GitHub itself is always changing, you would need to save a copy of [this wiki page](https://github.com/StylishThemes/GitHub-Dark/wiki/Highlighting) and add your tweaked theme to the bottom of the page to test it.

[...]

PR #568, opened 2017-11-30
Start to GitHub Material Theme

sparcut:
Regarding #522 

This is a **start** to the GitHub Material theme, works fine in the few languages I'd tried it in. However in others does not match at all - e.g. go between the GitHub and CodeMirror in CSS.

So this is **not** complete, but if anyone wants to pick it up, it should be pretty simple to complete.


the-j0k3r:
So is the GitHub portion of this complete?

silverwind:
Complete color matching between different syntax engines is not a requirement (as it is not possible). Source files should be split in three parts, `themes/src/{github,codemirror,jupyter}.css`. Jupyter theme is also missing.

So in summary, no, this is not quite ready yet.

silverwind:
No movement here since quite some time, if someone wants to continue this, raise a new PR.

PR #586, opened 2018-02-13
Fix color of repo description text

Poorchop:
The short repository descriptions have had dark text for a while now. In some cases, the text was white but more often than not, it was dark and unreadable.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -515,7 +515,8 @@
   .repository-meta .edit-repository-meta label, .access-token .token-description,
   .access-token .credential-authorization-menu .status-heading, .token-scope,
   .shelf-dismiss .close-button:hover, .team-listing .is-open .expand-nested-team,
-  .workspace-tree [aria-selected], .subheader-nav .nav-item.selected {
+  .workspace-tree [aria-selected], .subheader-nav .nav-item.selected,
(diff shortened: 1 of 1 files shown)


Poorchop:
I see it on almost every repo page. For some reason, this repo doesn't have that problem along with some others, but here is [one example](https://github.com/silverwind/droppy) along with how it looks on my end without the rule:

untitled

silverwind:
This sounds like a bug in your userstyle engine. Can you post the content of `<head>` when it happens?

Poorchop:
I'm using Stylish in Firefox 58.0.2.

>Both Stylus and the original Stylish move the injected stylesheet after the closing </body>, so this shouldn't be an issue.

The GitHub-Dark stylesheet appears before the body on the problem page but it shows up after `</body>` on this very page. I guess that means that the problem is with the add-on and not the selector. That doesn't really explain why I don't have the problem when I add the selector in this pull. In any case, it's probably not an issue with this userstyle.

silverwind:
A proper fix for these style races will come from https://github.com/openstyles/stylus/issues/248 with the `cssOrigin` option in Stylus.

Mottie:
> That doesn't really explain why I don't have the problem when I add the selector in this pull.

The selector you added has higher specificity.

PR #592, opened 2018-03-08
proposed fix for #591

### LINKED ISSUE #591: On hover expander dots styled vs unstyled border color.
Test URL https://github.com/sergi/jsftp/releases

* **Browser**: FF58.0.2
* **Operating System**:Win 10
* **Screenshot**:

dots

This video shows issue on styled, then how the unstyled displays the hover and finally my proposed fix to the styled on my local.

````html
<span class="main expander-text js-timeline-tags-expander">
  <span class="expander-dots">
    <span class="expander-dot"></span>
    <span class="expander-dot"></span>
    <span class="expander-dot"></span>
  </span>
  Show 4 newer tags
</span>
````
I also have a fix which is the last part of the video which is to add ```border-color: transparent !important;``` below https://github.com/StylishThemes/GitHub-Dark/blob/2747691085f9ad7bcf38e1579e496250dc32285e/github-dark.user.css#L1176 but am unsure if it breaks anything else, hence the issue and not a PR.

the-j0k3r:
fixes #591 

see #591

### DIFF
--- a/github-dark.user.css
+++ b/github-dark.user.css
@@ -1174,6 +1174,7 @@
   .select-menu-item.navigation-focus.selected, .result.selected, .result:hover,
   .CommunityTemplate-highlight, .diff-table button.add-line-comment {
     background: /*[[base-color]]*/ !important;
+    border-color: transparent !important;
   }
   .context-button:after {
     border-color: /*[[base-color]]*/ transparent transparent !important;


silverwind:
The rules where you added the border style is going to affect a lot more elements than those dotted buttons. Could you instead add your selector(s) [here](https://github.com/StylishThemes/GitHub-Dark/blob/2747691085f9ad7bcf38e1579e496250dc32285e/github-dark.css#L1481-L1487)?

silverwind:
Another option might be to set `border-color` to `/*[[base-color]]*/ #4183c4 !important;`. Oh, and please edit `github-dark.css`. The `github-dark.user.css` is autogenerated from it with the `grunt usercss` command.

the-j0k3r:
your suggestion also affects everything and Im not sure what end result is, so Ill leave it to you.

PR #604, opened 2018-04-04
Fixing text color on issue labels

jmona789:


Mottie:
Hi @jmona789!

Thanks for your contribution! Sadly, we can't apply an all-encompassing text color to issue labels. GitHub [calculates the color constrast](https://24ways.org/2010/calculating-color-contrast) based on the owner selected label background.

Within GitHub-Dark, we use the following selector to change the text color from `#000` to `#181818`:

```css
a.IssueLabel[href*="/issues?"][style*=" color: #0"], a.IssueLabel[href*="/pulls?"][style*=" color: #0"], .issue-card-label[style*=" color: #0"] {
  color: #181818 !important;
}
```

If you're seeing an issue with the color contrast, then please share the html so we can get a better idea of how to fix it.

jmona789:
Hi @Mottie, on this page: https://github.com/assemble/assemble.io/pulls the issue labels are getting the default link color due to them matching `a:not([href*="/labels/"])`

Mottie:
I'm not seeing the issue.

I do see that `a:not([href*="/labels/"])` is applied, but it is being overridden by the selector I shared above. Please make sure you're using the most up-to-date version of GitHub-Dark (currently 1.20.30).

silverwind:
They match on `a.IssueLabel[href*="/issues?"][style*=" color: #0"]` which is fine. I think this issue should be fixed by GitHub instead as these lables are black-on-blue without our style applied too. FTR, these inline styles are applied by GitHub:

````css
element {
    background-color: #207de5;
    color: #000000;
}
````

jmona789:
My apologies, I had not updated.  Closing this PR, thanks

PR #615, opened 2018-05-13
polls: improve contrast

the-j0k3r:
before
polls-b

after (contrast 50%)
polls-a

I think maybe a contrast of 70% feels better here tbh.

after (contrast 70%)
polls-a-70

let me add that instead

OK 90% is better is less subtle contrast in bg but slightly more than without contrast.

 #contrast 90%
polls-a-90

poll-u

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2185,7 +2185,7 @@
   /* invert network graph; gh-polls (https://github.com/apex/gh-polls) */
   div#network.network > div > canvas, a[href*="api.gh-polls.com"] img,
   img[data-canonical-src*="prod/timeseries/"], img[alt="FOSSA Status"] {
-    filter: invert(90%) hue-rotate(180deg) contrast(50%) !important;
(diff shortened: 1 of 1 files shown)


Mottie:
Hi @the-j0k3r!

Sorry, this was already fixed in https://github.com/StylishThemes/GitHub-Dark/commit/8f185877e633d9bde393fd6a03f92558b0660ab8... we'll push a new release soon.

Mottie:
Oh wait, you've added contrast... hmmm.

Mottie:
I'm not sure I like the brighter progress bar background... it's a lot more subtle when GitHub-Dark is turned off. Personally, I don't think it needs the contrast, but lets see what @silverwind thinks.

the-j0k3r:
yes it is more subtle but not as subtle as without, maybe a contrast of 90% is better?

the-j0k3r:
added contrast 90% and image to first post

PR #626, opened 2018-06-02
fix the CodeMirror-dialog-top (search bar) bottom border

the-j0k3r:
Wasn't sure where to place this snippet.

Before
before-fix
After
after-fix

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1109,10 +1109,6 @@
   .CodeMirror ::-moz-selection {
     background: rgba(255, 255, 255, .12) !important;
   }
-   /* styles the search bar bottom border */
-  .CodeMirror-dialog-top {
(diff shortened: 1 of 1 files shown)


Mottie:
Hi @the-j0k3r!

Thanks for your contribution! Instead of adding a whole new entry, please add the `.CodeMirror-dialog-top` selector to the end of [this block](https://github.com/StylishThemes/GitHub-Dark/blob/master/github-dark.css#L1634) that sets the `border-color` to `#484848`.

Mottie:
Thanks!

the-j0k3r:
>Thanks for your contribution! Instead of adding a whole new entry....

Youre very welcome for this drop of dew in the rainforest.

I thought as much about where to place it but hey :)

Issue #644, opened 2018-06-19
reopen #622 quick jump search bar is broken again


the-j0k3r:
just to make sure its not missed again in the rush of pushing a build out #622

Mottie:
Yes, I'm on that too... it looks like they completely removed `.navigation-focus` and replaced it with `[aria-selected="true"]`. Ugh... we'll probably need to maintain both sets of selectors so the GitHub Enterprise users won't have a problem.

What do you think of adding this to the generate script @silverwind?

the-j0k3r:
Im sure theres more... but this time Ill let you do your thing.

the-j0k3r:
the only things left is `.jump-to-suggestions-path .jump-to-octicon` is too dark on `#949494`

the-j0k3r:
What a mess that search menu jump to is... if you hover over selections after adding 

``css
 .jump-to-suggestions-path:hover .jump-to-octicon,
  .jump-to-suggestions-path:hover .jump-to-suggestion-name {
    color: #efefef !important;
  }
```
then move the mouse away the whole thing is unreadable and it knocks elements out while hovering I bet related to https://github.com/StylishThemes/GitHub-Dark/issues/632 cause those elipses in path while hovering is nasty.

Mottie:
This should be fixed in 1.20.49... I removed a bunch of duplicate selectors, but I might have been a bit too aggressive. Let me know if anything broke in the process.

the-j0k3r:
Sofar, arm chair. :) And I think I fixed the knockout while hovering the quick jump search thing.

PR #646, opened 2018-06-19
fixes: part fix hovering text/icon for #644

the-j0k3r:
before/after
hover

If I could figure out the knockout issue Ill add that as local fix


Mottie:
I've already fixed this issue... updating userstyles.org now. Let me know if you encounter any issues.

PR #650, opened 2018-06-19
issues: really fix #647 and prevent jump to from shifting to left.

### LINKED ISSUE #647: Knockout icons while hovering quick jump search menu.
* **Browser**: 66.0.2
* **Operating System**: W10pro 1709
* **Screenshot**:

Im not sure if this is me only because I have a 4K screen running at 2560x1440p or its general.
I suspect this is another upstream bug, but need confirming others also see the issue.

hover

The gif shows the issue before and after I applied a local fix.

````css
.jump-to-suggestions-path:hover .jump-to-suggestion-name.css-truncate-target {}
````

the-j0k3r:
max-width is wrong here see
capture

width is correct and jump-to is in right position.
capture

@Mottie did you not see my fix in userCSS? lol

@xt0rted see what I mean? Its bad enough that long titles are cut off but as is you get to see even less 
max, pff, like Pepsi, gimme Coke.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1884,7 +1884,7 @@
     max-width: 250px !important;
   }
   .jump-to-suggestions [aria-selected="true"] .jump-to-suggestion-name.css-truncate-target {
-    max-width: 50% !important;
+    width: 50% !important;
   }
   /* tooltip */
   .octotip {
--- a/github-dark.css
+++ b/github-dark.css
@@ -1884,7 +1884,7 @@
     max-width: 250px !important;
   }
   .jump-to-suggestions [aria-selected="true"] .jump-to-suggestion-name.css-truncate-target {
-    width: 50% !important;
+    width: 41% !important;
   }
   /* tooltip */
   .octotip {


Mottie:
> @Mottie did you not see my fix in userCSS?

I did, but the other definitions weren't necessary. And for some reason, I remembered seeing a `max-width` instead of just `width`. Anyway, even with 50%, there is an issue on an organization page with the "In the organization ↵" entry. You'd have to reduce the width to 41% to fix it.

Maybe we just live with the one pixel offset and save ourselves 123 bytes. The GitHub dark style is already huge enough.

the-j0k3r:
I dont care, Ill fix it locally but the max-width has to go.

Mottie:
If you think the 41% is too narrow, then lets just revert the fix.

the-j0k3r:
I cant tell the difference only that the shifting is cured.

the-j0k3r:
There you go. :) and more bugs I found too.

PR #652, opened 2018-06-20
style tree-browser chevron + text search match & sundries.

the-j0k3r:
ree-browser chevron and text search match are 2254198 and 43cc8fd 

chvreon


the-j0k3r:
With a2822a3  Im adding this the main text on a blue banner loacted at https://developer.github.com/ bottom page.

banner

With 07fb24c Im moving a couple of selectors to new position in the rule, this releases two lines (all bow to my might) :crown: .
I did this because there's many areas like this and could be improved in similar way. So @Mottie can decide it takes me two minutes to do the lot.

That said I can go two ways, either do a general cleanup of same nature or ignore it.
Can be reverted if not wanted.

github-dark.css
```diff
@@ -1091,7 +1090,9 @@
   table.files td.message a:hover, .link-gray:hover, .news .link-gray-dark:hover,
   .news a.link-gray-dark:hover, .reblame-link:hover,
   .orgs-help-item-octicon .octicon, span.btn-link, .btn-link, a.btn-link,
-  .merge-status-list .status-actions:hover {
+  .merge-status-list .status-actions:hover,
+  .tree-browser tr.navigation-focus td .octicon-chevron-right,
+  .tree-browser tr[aria-selected="true"] td .octicon-chevron-right, {
     color: /*[[base-color]]*/ #4183c4 !important;
```
Mottie: Was adding the extra `td` necessary?

github-dark.css
```diff
@@ -2787,18 +2788,16 @@
   .blog .text-delta, .blog .text-gamma, .blog h1, .blog h2, .blog h3, .blog h4,
   .blog h5, .site-header-nav a, .menu-item.selected, .overview .text-beta,
   article h2, #message-list h1, #message-list h3, .graph-page h1, .graph-page h3,
-  .dashboard-notice h2,
   .prose-diff.collapsed .rich-diff-level-zero.expandable:hover .octicon,
-  .counter, .Counter, .state, .State, a.state, a.State,
-  .site-header-dark .site-header .site-header-link,
+  .counter, .Counter, .state, .State, a.state, a.State, .dashboard-notice h2,
+  .site-header-dark .site-header .site-header-link, .version-picker a:hover,
   .markdown-format .octicon-link, .Box-row--hover-blue.link-gray:hover,
```
Mottie: Please add all new selectors to the end of the block, next to the `{`.

github-dark.css
the-j0k3r: If you look at the default css this is how it is, so I suspect this didnt work

github-dark.css
the-j0k3r: They are not new entries they were moved within the rule to cleanup.

Mottie: Yes, it makes the css pretty, but consistently adding new/modified selectors to the end helps us estimate the age of a selector. The class names in older selectors may not even exist anymore, then we can decide to remove them without much debate.

the-j0k3r: OK. Ill know for next time.

github-dark.css
```diff
@@ -1091,7 +1090,9 @@
   table.files td.message a:hover, .link-gray:hover, .news .link-gray-dark:hover,
   .news a.link-gray-dark:hover, .reblame-link:hover,
   .orgs-help-item-octicon .octicon, span.btn-link, .btn-link, a.btn-link,
-  .merge-status-list .status-actions:hover {
+  .merge-status-list .status-actions:hover,
+  .tree-browser tr.navigation-focus td .octicon-chevron-right,
+  .tree-browser tr[aria-selected="true"] td .octicon-chevron-right, {
```
silverwind: Trailing comma here invalidates the entire block.

the-j0k3r: fixed thanks ;) in between moving from stylus to git maybe it slipped into a coma :)

github-dark.css
Mottie: It wasn't working because they switched everything over to using `[aria-selected="true"]`. We're leaving in the `.navigation-focus` for GitHub Enterprise users since they lag behind in changes.

the-j0k3r:
@Mottie done, removed the text-gamma bits is there something else?

Mottie:
Nope, thanks!

Mottie:
@silverwind I was planning on a new release. Do you have anything in the works where'd I'd need to wait?

PR #660, opened 2018-06-24
fix: prereceive feedback banner fixes #658

### LINKED ISSUE #658: unstyled creating empty releases with not tag feedback banner
* **Browser**: ff66.02
* **Operating System**: win 10 pro 
* **Screenshot**:

So goto releases, create new, then just push the publish release with nothing filled in

capture

````html
<div class="prereceive-feedback">
  <h3 class="prereceive-feedback-heading">
    Tag could not be created. Pre-receive hooks failed.
  </h3>
  <pre></pre>
</div>
````

the-j0k3r:
capture

The pre clearly will come into use with some more specific msg later, see issue ticket for before fix looks.

this styles the banner and pre since the .prereceive-feedback-heading` is already handled by autogenerate.
https://github.com/StylishThemes/GitHub-Dark/blob/8313559707f85deb49829f033eecc0f5e554fb76/github-dark.css#L990

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -3222,7 +3222,7 @@
   /* yellow text */
   /* .bg-pending might be a GitHub bug as it sets the fg */
   .text-pending, a.text-pending, .text-renamed, a.text-renamed, .bg-pending,
-  a.bg-pending, .prereceive-feedback pre {
+  a.bg-pending {
     color: #cb4 !important;
   }
   /* yellow border (travis) */
@@ -4117,7 +4117,7 @@
   .graphiql-ide.signed-out:after {
     background: #000 !important;
   }
-  .graphiql-container {
+  .graphiql-container, .prereceive-feedback-heading {
     color: #ddd !important;
   }
   .graphiql-container .docExplorerShow:before {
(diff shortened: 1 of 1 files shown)


the-j0k3r:
@Mottie what about this one :)
dance

Mottie:
Hi @the-j0k3r!

Hmm, what do you think of the urine-yellow color? Would that look better white? What do you think?

the-j0k3r:
I was inclined to use #ddd but because this actually has a wider left border I left it like this until feedback came.

the-j0k3r:
Ah yes, the reason why its yellow its causes its a autogenerated thingy

the-j0k3r:
I moved `.prereceive-feedback-heading` from the *urine*-yellow autogenerated bit and added it to #ddd along with `.prereceive-feedback pre`
 👍

Mottie:
Don't bother removing selectors from the auto-generated sections, it'll just be added back the next time we regenerate. Also, you don't need to add the `.prereceive-feedback pre` selector since it's already white - add some content to it and you'll see.

the-j0k3r:
There.

# Is that better

edgar

the-j0k3r:
Thanks for tips and bits and pieces :+1:

PR #661, opened 2018-06-28
invert 1clr-code-hosting svg

the-j0k3r:
1clr-code-hosting

Must be logged off and on https://github.com/ to see this page and image.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2325,7 +2325,7 @@
   g-emoji[fallback-src$="1f43e.png"], g-emoji[fallback-src$="1f43e.png"],
   img[src*="resources-logo"], svg.IconNav-img,
   /* https://github.com/ten */
-  .container-xl[id] svg {
+  .container-xl[id] svg, body.logged-out .col-md-7 svg {
     filter: invert(83%) !important;
   }
   .container-xl[id] img {


Mottie:
Hi @the-j0k3r!

The `svg.mr-4` is way too generic.

the-j0k3r:
IDK thats the actual image class, what do you suggest instead?

Mottie:
This is way too much trouble I think... To specifically target that svg, you'd have to do something like this:

```css
@-moz-document regexp("^https:\\/\\/github\\.com\\/?$") {
  body.logged-out .col-md-7 svg {
    filter: invert(80%) !important;
  }   
}
```

Did I mention GitHub-dark is already way too big? So I'm inclined to ignore this one icon on the main page only seen while logged out. What do you think @silverwind?

the-j0k3r:
Same as the other one...

blow-brains

Mottie:
You didn't have to close this PR... I said I was inclined, not totally against it. Anyway, maybe we don't need to wrap that selector in a `@-moz-document`... the `body.logged-out .col-md-7 svg` may be specific enough.

the-j0k3r:
I think I pressed wrong button there, anyway updated and here it is.

silverwind:
>  ignore this one icon on the main page only seen while logged out

If it's on the "main" site, I guess we should have it. There might be users who manually log in each time.

Overall, we need to start eliminating old selectors. I estimate that at least 50% of our selectors are no longer in use. I'd start by cutting big block of selectors and then fix any issues that come up.

the-j0k3r:
Why dont you guys dump the image data URI links and host them via cdn/rawgit from the repo? that would save a couple of bytes, it may not be as fast but it probably wouldn't hurt that much. Or no?

silverwind:
I prefer datauris because data is guaranteed to be available instantly without delay. Depending on how these images are cached, loading them from an URL can introduce loading delays.

the-j0k3r:
Sure, I understand the reason, but if there is an URL to an actual image/png/svg whatever on GH servers isn't it the same thing?

You have a few checkboxes for instance, I haven't checked but if there are then IDK what the difference is, except local always loads faster but you dont need checkboxes to load faster than the page renders.

the-j0k3r:
Ah never mind theres non checkmarks and using `content` doent work as well.

Issue #607, opened 2018-04-10
10 year celebration page looks really bad


xt0rted:
* **Browser**: Chrom 65
* **Operating System**: Win10

https://github.com/ten

Mottie:
Hi @xt0rted!

Thanks! I'll get it fixed... err excluded; too many dark background images to bother.

xt0rted:
I just released my first style on userstyles.org, when you make a change like this do you manually update the site or is that automated?

Mottie:
Manually update :(

xt0rted:
Now I feel bad for reporting all those issues 🙁

Mottie:
Naw... we did ask to have [auto-updates from GitHub](https://github.com/stylish-userstyles/userstyles/issues/50), but nothing ever happened.

xt0rted:
I think this change broke some urls, for instance this page isn't styled now but it was an hour ago https://github.com/dncuug/X.PagedList/blob/master/src/X.PagedList/PagedListExtensions.cs

I think it's because of _.../PagedListEx**TEN**sions.cs_

xt0rted:
Here's another that's broke https://github.com/Azure/azure-functions-durable-extension

Mottie:
DOH

Mottie:
Ugh... well, I think we'll just have to deal with it.

I can invert the darker background images with these *ugly* and inefficient selectors:

```css
[class*="TenYear"] svg, [class*="TenYears"] ~ div svg, svg[class*="TenYear"],
  img[src*="tenyear"] ~ svg {
  filter: invert(100%) !important;
}
```

But there isn't much I can do with the humongous image at the top.

Mottie:
Actually, this is better:

```css
/* ==UserStyle==
@name        github.com/ten
@namespace   StylishThemes
@version     1.0.0
@homepageURL https://github.com/StylishThemes/GitHub-Dark/issues/607
@license     CC-BY-SA-4.0
@author      StylishThemes
==/UserStyle== */
@-moz-document url("https://github.com/ten") {
  svg {
    filter: invert(100%) !important;
  }
  img {
    z-index: 100 !important;
  }
}
```

the-j0k3r:
@Mottie naturally fitting when white images burn retinas.

![10-years](https://user-images.githubusercontent.com/31389848/42033150-ccc2adbc-7ad3-11e8-9561-28c4dbad2148.gif)

```css
[...]

PR #662, opened 2018-06-28
10-years: burn white images

the-j0k3r:
Related to #607

<s>Also</s> Handles https://github.com logged off 10 white image

capture

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2331,6 +2331,9 @@
   .container-xl[id] img {
     z-index: 100 !important;
   }
+  img[src$="10.png"] {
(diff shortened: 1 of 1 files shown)


the-j0k3r:
@Mottie let me see if I find a better selector for the single 10 year image

Mottie:
I'd use `body.logged-out img[src$="tenyears/10.png"]`, but as I said, I wouldn't use `color-burn` here because it all depends on the user selected background image.

the-j0k3r:
The alternative to use filters would not be much better in that case.

blow-brains

PR #663, opened 2018-06-30
data uri: squeeze off the fat

the-j0k3r:
Its not much, maybe not even worth it, but fat trimmed it was.


auscompgeek:
I suspect these might not need to be base64-encoded either. They're just text after all, right?

the-j0k3r:
its svg so yes, its less size if its not base64

the-j0k3r:
**before** 222 KB (227,582 bytes)
**after** 221 KB (227,243 bytes)

silverwind:
This is just running them through `svgo`, right? I think you can actually drop the XML tag too for further reductions, but I'm not 100% sure. Best to test this in a few browsers.

> its less size if its not base64

One can actually represent SVG unencoded but every browser parses these strings differently (last I checked, Firefox needs percent-encoding for characters that are not url-safe while Chrome does not), so base64 was chosen for best compatibility.

the-j0k3r:
>This is just running them through svgo, right? 

Not svgo only, http://b64.io/ alone offered better results in some cases just doing a straight base64 to svg and then passed via http://b64.io/.

>Best to test this in a few browsers.

Sure test away. Ill do same and report back.

>One can actually represent SVG unencoded but every browser parses these strings differently 

Some places directly link to a svg, depends on its lace of use  is what I thought otherwise everyone would base64 encode all svgs and they dont.

silverwind:
Sample here: https://jsfiddle.net/c0ak9pq1/10/

The second case with no base64 does not work in Firefox but does in Chrome and Safari. Also xmlns seems to be required unfortunately.

the-j0k3r:
Well idk what to say, this seems to work around FF Chrome and Opera browsers and files have been trimmed in one case quite considerably.

Up to you.

auscompgeek:
Hm. Looks like it's only `#` which needs URL encoding for Firefox, since it'd be interpreted as the fragment otherwise. I don't expect any other character to be considered special in data URIs though…

Mottie:
Hmm, you're considering 339 bytes as "fat"? I hope you're weren't ever a drill instructor or dietitian...

the-j0k3r:
> Hmm, you're considering 339 bytes as "fat"?...

339 less bytes is certainly less chewing, 

 >I hope you're weren't ever a drill instructor or dietitian

Hope, like Bob and Uncle are a good staple diet for this.

silverwind:
@auscompgeek nice find with the `#`. I can confirm it works in Firefox, Safari and Chrome:

https://jsfiddle.net/c0ak9pq1/16/

Would be great if someone with access to Edge can check if below example shows two black boxes with checkmarks.

If it does work, I think we should go with just utf8 over base64 which brings a 33% size reduction. It also has the benefit of easy editing.

the-j0k3r:
> Would be great if someone with access to Edge can check if above example shows two black boxes with checkmarks.

I try to avoid Edge for anything (and haven't uninstalled it yet since upgrade to 1709), but since its for a good cause...

capture

silverwind:
Nice! Please change this PR to use the utf8 data uris.

By the way, relevant Firefox bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1312700.

the-j0k3r:
> By the way, relevant Firefox bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1312700.

Another bug for the ages.

silverwind:
> Another bug for the ages.

In an effort to stay relevant, Mozilla has been doing changes for the sake of web compatibilty (They event went as far as implementing `-webkit-` CSS prefixes), maybe we'll get a fix.

the-j0k3r:
**ORIGINAL** 222 KB (227,582 bytes)
**With  SVG** 221 KB (226,546 bytes)
**Previous base64 squeeze** 221 KB (227,243 bytes)

I think we can remove even more from this. I just took the SVGs as they were, so bare with me.

Mottie:
> ORIGINAL 222 KB (227,582 bytes)
> With SVG 221 KB (226,546 bytes)

Still too fat!! 😉

the-j0k3r:
> Still too fat!! wink

Your time to drill :)

I'll have a better offer as soon as I can with extra extra optimized SVG's but if someone beats me to it :) Bobs your uncle.

the-j0k3r:
@Mottie with the extra trimming its now **221 KB (226,364 bytes)** mmm interesting that with my original base64 proposal it was smaller yet :) and hell no, Im not putting that back :)

Still less bytes is less chewing

Let me know and Ill squash these commits after you have tested it only adding it so you can visully see the differences maybe spot something I missed and if you prefer either version its an option before I force push it to oblivion.

:+1:

PR #664, opened 2018-06-30
readme updates and badges.

the-j0k3r:


README.md
```diff
@@ -24,17 +24,19 @@
...
-    * or, add it [manually](https://raw.githubusercontent.com/StylishThemes/GitHub-Dark/master/github-dark.css) into the editor.
-      * Use the [grunt build process](https://github.com/StylishThemes/GitHub-Dark/wiki/Build) to customize your GitHub Dark theme.
-      * Please refer to the [installation documentation](https://github.com/StylishThemes/GitHub-Dark/wiki/Install) for more details.
+#### If you're using a browser extension:
+* Stylus - get the addon for [Firefox](https://addons.mozilla.org/en-US/firefox/addon/styl-us/), [Chrome](https://chrome.google.com/webstore/detail/stylus/clngdbkpkpeebahjckkjfobafhncgmne) and [Opera](https://addons.opera.com/en-gb/extensions/details/stylus/).
+* Stylish - get the addon for [Firefox](https://addons.mozilla.org/en-US/firefox/addon/2108/), [Chrome](https://chrome.google.com/extensions/detail/fjnbnpbmkenffdnngjfgmeleoegfcffe), [Opera](https://addons.opera.com/en/extensions/details/stylish/), [Safari](http://sobolev.us/stylish/) and [Firefox Mobile](https://addons.mozilla.org/en-US/firefox/addon/2108/).
+* FreeStyler is available for [Chrome](https://chrome.google.com/webstore/detail/freestyler/hihigldmabkodfpehkgdemjklmaebmca).<br>
```
silverwind: You could just do `$extension for Firefox, Chrome, ...`

README.md
```diff
@@ -2,19 +2,19 @@
...
+  <a href="https://github.com/StylishThemes/GitHub-Dark/network">
+    <img src="https://img.shields.io/github/forks/StylishThemes/GitHub-Dark.svg?style=for-the-badge" alt="Fork">
   </a>
   <a href="https://david-dm.org/StylishThemes/GitHub-Dark?type=dev">
-    <img src="https://img.shields.io/david/dev/StylishThemes/GitHub-Dark.svg?label=%20devDependencies%20" alt="devDependencies">
+    <img src="https://img.shields.io/david/dev/StylishThemes/GitHub-Dark.svg?label=devDependencies&style=for-the-badge" alt="devDependencies">
   </a>
   <a href="https://gitter.im/StylishThemes/GitHub-Dark">
-    <img src="https://img.shields.io/gitter/room/StylishThemes/Github-Dark.js.svg?maxAge=2592000" alt="Gitter">
+    <img src="https://img.shields.io/gitter/room/StylishThemes/Github-Dark.js.svg?maxAge=2592000&style=for-the-badge" alt="Gitter">
```
silverwind: I think I prefer the old (smaller) badges.

README.md
the-j0k3r: I just wanted to improve the easier to follow factor without too much nesting

README.md
the-j0k3r: Let me try the smaller flat-square instead, but sure I thought these may not be that attractive at first look.

ddavison: getting rid of my badges, huh :) @the-j0k3r

Mottie: I'd prefer to leave the badges as-is...

the-j0k3r: @ddavison actually I would prefer your badge for starring, but your styles are limited to two styles only, could you add more?.

@Mottie sure, OK, I only tried to make them all look the same instead of a mix match of different styles

the-j0k3r:
@silverwind @ddavison @Mottie see before and after

if @ddavison added a flat square style then we could have a good star count + all badges look similar and not mix and match.

badges

I can remove the commit of course.

the-j0k3r:
Badges are more or less back to normal wiht just enough updates to make all but one consistent.

Only one left which I prefer for accurate star count is @ddavison but the font is much bigger in some setups with FF due to some bug that was fixed in shieldsIO thast not in his repo.

README.md
```diff
@@ -2,19 +2,19 @@
   <img alt="githubdark-logo" src="https://rawgit.com/StylishThemes/logos/master/github.dark/githubdark-mini.svg" width="580">
   <br>
   <a href="https://github.com/StylishThemes/GitHub-Dark/tags">
-    <img src="https://img.shields.io/github/tag/StylishThemes/GitHub-Dark.svg?label=tag" alt="Tag">
+    <img src="https://img.shields.io/github/tag/StylishThemes/GitHub-Dark.svg?label=version&style=flat" alt="Version">
   </a>
   <a href="https://github.com/StylishThemes/GitHub-Dark/stargazers">
-    <img src="http://github-svg-buttons.herokuapp.com/star.svg?user=StylishThemes&repo=GitHub-Dark&style=flat&background=007ec6" alt="Star">
+    <img src="http://github-svg-buttons.herokuapp.com/star.svg?user=StylishThemes&repo=GitHub-Dark&style=flat&background=007ec6" alt="Stars">
```
silverwind: I don't suppose we have a "stars" badge available somewhere? This singular always bugged me.

the-j0k3r: Sorry not that Im aware.

The singular along with font fixes is up to @ddavison, on the stars badge, a suggestion, he could add a ?label=stars override or whatever similar to what shieldsIO does

I could use the shieldsIO stars badge (and use plural) but I dont like the rounded way they count 2896 = 3K, may as well invent a more arbitrary number like 9999999 stars, hahaha

PR #679, opened 2018-07-06
Fix forks' network tree spacing

tambry:
Before:
2018-07-06_13-46-18_chrome

After:
2018-07-06_13-44-35_chrome

Default:
2018-07-06_13-46-37_chrome

Traffic before:
2018-07-06_14-49-28_chrome

Traffic after:
2018-07-06_14-54-07_chrome

PR #681, opened 2018-07-06
issues: tree view fixes re: #679

the-j0k3r:
Test these out @tambry 

gif shows initially zoom at 100% and each full scroll into oposite direction increases 10% zoom up to 130% zoom with no gaps.

Browser FF 61.0
 
zoom 100% initial scroll to bottom
zoom 110% scroll to top
zoom 120% scroll to bottom
zoom 130% scroll to top

the gif is 15MB (squeezed from original 37MB) so download it...

https://www.dropbox.com/s/5pypt47ab9albi6/treeview.gif?dl=1

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -3447,10 +3447,11 @@
   }
   /* network members */
   .network-tree {
-    width: 26px !important;
+    width: 19px !important;
     height: 0 !important;
     padding-top: 26px !important;
     background-repeat: no-repeat !important;
+    background-position: -2px center !important;
   }
   .network-tree[src$="t.png"] {
     background-image: url("data:image/gif;base64,R0lGODlhFAAaAIAAAMzMzP///yH5BAEAAAEALAAAAAAUABoAAAIkjI+pG8APY5O0uorfzRzt3n1g5pSTOTJiSq1s5L6ajMW0YgcFADs=") !important;


tambry:
I don't think it improves anything. In fact the gap between the dotted pictures increases and becomes more obvious. I think my original fix, but with the now pre-inverted backgrounds would be better. I'll make a PR for that.

Before:
2018-07-07_13-01-25_chrome

After:
2018-07-07_13-02-10_chrome

the-j0k3r:
> I think my original fix, but with the now pre-inverted backgrounds would be better. I'll make a PR for that.

that results in broken background in Firefox as well.

Mottie:
What if we reduced the height of the image? Or set the background size to cover?

the-j0k3r:
> What if we reduced the height of the image

I reduced the heigh of the image here physically to 25px from 26px removing the gap at top edge between canvas and dot, so perhaps reduce it also in CSS? let me push that in here.

the-j0k3r:
`background-size: cover !important;` screws with the center alignment and any more editing to files itself will be a very deep rabbit hole/can of worms (from experience with this issue in other projects), so I will bow out of that scenario.

Open to other suggestions though.

the-j0k3r:
`	background-position: -2px center !important;` is interesting.

popover

@tambry I reverted gifs to latest default and added tweak above, I wonder if that has a positive effect on the issue you see despite it just aligning the image  horizontally.

the-j0k3r:
Im out of ideas ATM so any suggestions welcome.

the-j0k3r:
> Looks good enough to me.

so I pushed width changes in, any other changes pls make some suggestion re: gaps I cant see any in FF

Mottie:
Are we all happy with these tweaks? I'm a bit under the weather, so I'll let you guys/gals argue about the minor details.

the-j0k3r:
>Are we all happy with these tweaks? I'm a bit under the weather, so I'll let you guys/gals argue about the minor details.

rahrahrah, argue argue argue, Im foine. here

the-j0k3r:
Thanks @tambry

PR #684, opened 2018-07-07
Make network tree more consistent with the default

tambry:
The spacing is now consistent with the default.

Before:
2018-07-07_13-17-16_chrome

After:
2018-07-07_13-18-06_chrome

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -3446,20 +3446,14 @@
     padding: 32px !important;
   }
   /* network members */
-  .network-tree {
-    width: 26px !important;
-    height: 0 !important;
-    padding-top: 26px !important;
-    background-repeat: no-repeat !important;
-  }
   .network-tree[src$="t.png"] {
-    background-image: url("data:image/gif;base64,R0lGODlhFAAaAIAAAMzMzP///yH5BAEAAAEALAAAAAAUABoAAAIkjI+pG8APY5O0uorfzRzt3n1g5pSTOTJiSq1s5L6ajMW0YgcFADs=") !important;
+    content: url("data:image/gif;base64,R0lGODlhFAAaAIAAAMzMzP///yH5BAEAAAEALAAAAAAUABoAAAIkjI+pG8APY5O0uorfzRzt3n1g5pSTOTJiSq1s5L6ajMW0YgcFADs=") !important;
   }
   .network-tree[src$="i.png"] {
-    background-image: url("data:image/gif;base64,R0lGODlhFAAaAIAAAMzMzP///yH5BAEAAAEALAAAAAAUABoAAAIhjI+pG8APY5O0uorfzRzt3n1gJo6WGaJcqUJsy7ywIgcFADs=") !important;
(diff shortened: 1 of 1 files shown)


the-j0k3r:
Like I said already, this breaks in Firefox but now instead of black like in https://github.com/StylishThemes/GitHub-Dark/pull/679#issuecomment-403027881 it goes white.

capture

the-j0k3r:
This also works in FF png instead of gif, IDK

```
  .network-tree {
    width: 26px !important;
    height: 0 !important;
    padding-top: 26px !important;
    background-repeat: no-repeat !important;
  }
  .network-tree[src$="t.png"] {
    background-image: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAaCAMAAACAXYxPAAAABlBMVEX////MzMw46qqDAAAAAXRSTlMAQObYZgAAABVJREFUeAFjwA8YEcyRpxIBR3IoAQAUggATwVxlNQAAAABJRU5ErkJggg==") !important;
  }
  .network-tree[src$="i.png"] {
    background-image: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAaCAMAAACAXYxPAAAABlBMVEX////MzMw46qqDAAAAAXRSTlMAQObYZgAAABJJREFUeAFjwA8YEcyRqXJUJQAPfQAOb0+bEgAAAABJRU5ErkJggg==") !important;
  }
  .network-tree[src$="l.png"] {
    background-image: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAaCAMAAACAXYxPAAAABlBMVEX////MzMw46qqDAAAAAXRSTlMAQObYZgAAABRJREFUeAFjwA8YEcwRqBIBRzAAABHKAA3ckBEiAAAAAElFTkSuQmCC") !important;
}
```

capture

I think as is its as good as its gonna get unless @tambry has another trick up his sleeve :)

tambry:
Crap, seems like it does. I thought the breakage in Firefox was only due to the inversion.

Seems like the problem is due using the `content` attribute. This seems to be [bug 714507 in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=714507).

Mottie:
Yeah, sadly using `content` doesn't work in Firefox. I'm going to close this PR... we can continue this discussion in #679.

the-j0k3r:
> I thought the breakage in Firefox was only due to the inversion.

sadly no... 

> his seems to be bug 714507 in Firefox.

This is one aspect of Firefox that seems consistent, there's many of these bugs At least 3 (oof the top of my head) that we track related to and cof "features" being tracked in relation to GitHub Dark.

🤟 bugs forever :sheep:

tambry:
I actually [asked a question about this on SO](https://stackoverflow.com/q/51223623/2176813). It seems it might not be a bug in Firefox afterall, as `content` is only supposed to apply to `::before` and `::after`. Using `::after` works in the sample in the SO question on Firefox, but not on Chrome and Edge.

silverwind:
> content is only supposed to apply to ::before and ::after

That is my understanding too.

PR #692, opened 2018-07-15
fix: button flash-close color

the-j0k3r:
flash-close

found while hunting for another issue to fix.


the-j0k3r:
> Better fix in e75460a, might have broken some buttons with that, but we can fix those later.

lol :blow my brains out. Thats the opposite definition of a better fix, simply because it did break other x buttons, I was trying to avoid wasting time reporting the bugs or fixing the issues this _fix_ broke.

silverwind:
It's best to mimic what is in GitHub's styles. If there is no `color` on `button` there, we shouldn't have one either. Long term, this will result in less issues.

Issue #674, opened 2018-07-04
Staff badge background is dark and hard to see now


xt0rted:
I noticed this a few days ago but I'm not sure if it was an intentional change in color, or if GitHub made a change that caused it. In the hover card it's not too bad, but on the profile page it seems unnoticeable at normal size.

* **Browser**: chrome
* **Operating System**: win10
* **Screenshot**:

![image](https://user-images.githubusercontent.com/831974/42290150-861a3e76-7f91-11e8-87f7-985f32ed4c4a.png)

![image](https://user-images.githubusercontent.com/831974/42290162-93f1fbc4-7f91-11e8-9e36-6ac3d37ac3f6.png)

* **HTML of the section where the issue occurs**:

The background color is coming from the `.bg-blue` rule

```html
<div class="flex-self-start">
  <div class="rounded-1 overflow-hidden">
    <a data-octo-click="hovercard-link-click" data-octo-dimensions="link_type:profile" href="/niik">
      <img class="d-block" height="60" width="60" alt="@niik" src="https://avatars2.githubusercontent.com/u/634063?s=88&amp;v=4">
    </a>
    <div class="d-block f6 bg-blue no-underline text-bold text-uppercase text-white text-center">Staff</div>
  </div>
</div>
````

````html
<ul class="vcard-details border-top border-gray-light py-3">
  <li class="vcard-detail pt-1 css-truncate css-truncate-target">
    <svg aria-label="Organization" class="octicon octicon-organization" viewBox="0 0 16 16" version="1.1" width="16" height="16" role="img"><path fill-rule="evenodd" d="M16 12.999c0 .439-.45 1-1 1H7.995c-.539 0-.994-.447-.995-.999H1c-.54 0-1-.561-1-1 0-2.634 3-4 3-4s.229-.409 0-1c-.841-.621-1.058-.59-1-3 .058-2.419 1.367-3 2.5-3s2.442.58 2.5 3c.058 2.41-.159 2.379-1 3-.229.59 0 1 0 1s1.549.711 2.42 2.088C9.196 9.369 10 8.999 10 8.999s.229-.409 0-1c-.841-.62-1.058-.59-1-3 .058-2.419 1.367-3 2.5-3s2.437.581 2.495 3c.059 2.41-.158 2.38-1 3-.229.59 0 1 0 1s3.005 1.366 3.005 4z"></path></svg>
    <span class="p-org" itemprop="worksFor">GitHub</span>
[...]

PR #700, opened 2018-07-19
marketplace app page and GH staff cards fixes,

### LINKED ISSUE #699: marketplace app pages styling look ugly.
* **Browser**: FF 66.0.1
* **Operating System**: Win10pro/Linux 
* **Screenshot**:

marketplace

e.g https://github.com/marketplace/buddy

- [x] font color on the right list is not really white `filter-item.selected` and friends aka `h4` and `.f3` are not styled?
- [x] the top border on the right box background is being overridden by `.bg-blue:not(.text-bold)`

````html
<li>
      <a href="/marketplace/buddy/plan/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW45MDM=#pricing-and-setup" class="filter-item ws-normal d-flex flex-justify-between p-2 p-md-3 mb-0 selected" data-pjax="" data-pjax-preserve-scroll="">
        
<div class="flex-auto flex-self-center">
  <h4 class="f3 d-inline-block marketplace-plan-emphasis">Free
  </h4>
  <p class="text-small mb-0 pr-2">1x concurrency, 1 GB RAM </p>
</div>
  <span class="f3 float-right marketplace-plan-emphasis">$0</span>

      </a>
    </li>
````

````html
<div class="position-absolute top-0 left-0 width-full bg-blue pt-1"></div>
````

the-j0k3r:
Tested in FF 66.0.1 Windows 10/Linux and G Chrome Windows

This fixes #699 and reverts 0937215 which should fix issue 2 in #699 and unbreak the fix for #674

marketplace


silverwind:
Thanks. Only part I was unsure about is if it's really necessary to define all the child elements `.filter-item.selected`, but I assume you had your reasons :)

the-j0k3r:
If you remove those the selected text wont be styled properly, maybe there is a magix selector that encompases all of those instead, if so, go for it.

silverwind:
There's probably another wrong rule that styles them wrong, but I'm too lazy right now to investigate :)

the-j0k3r:
;) I hear you. 

I looked and aside from autogenerated stuff which seems in right places, I tried removing the manual ones to no good result. Still I could have missed some other override, but it wasnt plainly obvious to me.

That said, I have no doubts that a cleaner solution may exist, but I was equally lazy trying more permutations.

Issue #598, opened 2018-03-23
Floating header in PR files view covers selected line


maxrothman:
In the PR files view (e.g. `https://github.com/<owner>/<repo>/pull/<number>/files`), when a line has been selected, the floating file header covers the selected line.

* **Browser**: Chrome
* **Operating System**: MacOS 10.13.3

Steps to reproduce:
* Go to the files view for a PR with a relatively large change (several pages of scroll)
* Click a line. Note the URL changes to something like `https://github.com/<owner>/<repo>/pull/<number>/files#diff-<sha>-L<num>` and the view jumps to the selected line.

Expected result:
<img width="1169" alt="screen shot 2018-03-23 at 1 03 43 pm" src="https://user-images.githubusercontent.com/2607086/37843247-c7d07bca-2e9a-11e8-9975-ac39dfeb049d.png">
Actual result:
<img width="1214" alt="screen shot 2018-03-23 at 1 04 08 pm" src="https://user-images.githubusercontent.com/2607086/37843262-ce1f3b38-2e9a-11e8-896c-0c415e8b648a.png">

* **HTML of the section where the issue occurs**:
````html
<tr>
  <td id="diff-4e08e0b4eed84c9a9d19372de85fa474aL3" data-line-number="3" class="blob-num blob-num-deletion js-linkable-line-number selected-line"></td>
  <td class="blob-num blob-num-deletion empty-cell selected-line"></td>
  <td class="blob-code blob-code-deletion selected-line">
    <button class="btn-link add-line-comment js-add-line-comment js-add-single-line-comment" data-path="src/Cron.dockerfile" data-anchor="diff-63f2befbc4fcc350af2072e9060ca77a" data-position="3" data-side="left" data-line="3" type="button" aria-label="Add line comment">
      <svg class="octicon octicon-plus" viewBox="0 0 12 16" version="1.1" width="12" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M12 9H7v5H5V9H0V7h5V2h2v5h5z"></path></svg>
    </button>
    <span class="blob-code-inner">-<span class="pl-k">RUN</span> apt-get update &amp;&amp; apt-get -y install cron</span>
  </td>
</tr>
````
This problem occurs because when the anchor is added to the URL (via a `click` event handler) the view jumps to put the element at the top of the screen. Since the floating file header is also at the top of the screen, it covers the line.

[...]

PR #666, opened 2018-07-03
re-tweak: issue #603 and move to grey bg fixes #665

the-j0k3r:
I think the blue color may look more natural than the brown, which just looks out-of-place.

Im sure there must be some very good reason to why GitHub-Dark chose to address issues caused by refined GitHub well off default design choices that are not part of default GitHub stylesheet.
But since that was the case, I propose a more suitable color that is actually present around comments, like the blue.

Idk if a border is needed and definitely can be tested as soon as this is triggered.

PR #701, opened 2018-07-19
Pull Request: Add specific support for refined-github stickied headers

silverwind:
Another approach to https://github.com/StylishThemes/GitHub-Dark/issues/598. This basically removes our opinionated stickying and supports the 60px high stickied header of `refined-github` specifically. I think this is a good compromise, if users want stickied filenames, they can install that extension.

Example with refined-github:

screenshot 2018-07-19 at 20 32 23

And without:

screenshot 2018-07-19 at 20 33 09

@the-j0k3r

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1295,13 +1295,6 @@
     position: -webkit-sticky !important;
     top: 8px !important;
   }
-  /* copied from Refined GitHub extension */
-  .pull-request-tab-content .diff-view .file-header {
-    position: sticky !important;
-    position: -webkit-sticky !important;
-    top: 39px !important;
-    z-index: 6 !important;
-  }
   /* User time line firsts */
   img[src$="profile-joined-github.png"] {
     padding-bottom: 20px !important;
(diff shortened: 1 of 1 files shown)


silverwind:
Thought I'm not sure which way to go. On one side, I find the stickied filenames quite useful and I'd accept them breaking the selected line scroll slightly. On the other hand I do not want to install `refined-github` with all the stuff that comes with it.

the-j0k3r:
Eh IDK what to tell you, I was getting used to having it stuck (with the bug exception) and I definitely dont want to use refined GH as its too opinionated and not everything makes sense the way they do it.

Bottomline: I dont see the selected line being solved any better with RGH and without its obviously a non issue.

The second proposed on comment on #598 further reveals more pixels but if you look too closely when youre scrolling you see it one pixel on the margin may be too obvious. Still the best compromise IMO.

xt0rted:
These sticky headers are a must when reviewing PRs but I don't want some of the really opinionated things in `refined-github` that I can't turn off. I don't care about the highlighted line being covered (only 1 person has ever mentioned it anyhow). If this is removed from `github-dark` I'll just end up adding it to my custom style which also fixes the change in #666.

the-j0k3r:
> I don't care about the highlighted line being covered 

You may not, I do.

silverwind:
Guess no one here really likes `refined-github` 😆 Let's close this one, we'll keep the sticky header.

Mottie:
@xt0rted You *can* turn off stuff in refined GitHub... but I don't use it either LOL

xt0rted:
Maybe I'll install it again then, but I'm sure something annoying won't be easily disabled and I'll uninstall it by the end of the day.

the-j0k3r:
Thats what I found when trying RGH, something distracting and not that useful cant be disabled and ends up getting on my nerves.

xt0rted:
`refined-github` hasn't been too bad now that I can disable features. It's lasted 27 hours so far, which is 26 more than last time I tried it 🤪

I just wish they provided a way to turn off (some of) their css-only things.

**Disabled so far**
```
monospace-textareas
move-marketplace-link-to-profile-dropdown
open-all-notifications
add-trending-menu-item
move-marketplace-link-to-profile-dropdown
add-diff-view-without-whitespace-option
add-readme-buttons
mark-merge-commits-in-list
```

Issue #702, opened 2018-07-20
Clear search button/icon is hard to see


xt0rted:
It's been awhile since I've looked at this, so I'm not sure if the color changed recently, but I thought this was more visible than it is now.

* **Browser**: chrome
* **Operating System**: win10
* **Screenshot**:

![image](https://user-images.githubusercontent.com/831974/42977288-4c8948aa-8b94-11e8-864b-78ef2a30f864.png)

* **HTML of the section where the issue occurs**:

````html
<div class="issues-reset-query-wrapper">
    <a class="issues-reset-query" href="/StylishThemes/GitHub-Dark/issues?q=is%3Aissue+is%3Aopen">
      <svg class="octicon octicon-x issues-reset-query-icon" viewBox="0 0 12 16" version="1.1" width="12" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M7.48 8l3.75 3.75-1.48 1.48L6 9.48l-3.75 3.75-1.48-1.48L4.52 8 .77 4.25l1.48-1.48L6 6.52l3.75-3.75 1.48 1.48L7.48 8z"></path></svg>
      Clear current search query, filters, and sorts
</a>  </div>
````

Mottie:
Hi @xt0rted!

Thanks for reporting this issue... it's fixed in the master branch & it'll be available in the next update.

xt0rted:
@Mottie I installed `refined-github` again to try it out with some of the annoying things disabled. It's adding commit links to the status labels above. With `github-dark` disabled they're white, but with it enabled they're blue.

Is this something you want to try fixing? I'm probably going to turn that feature off since there's already a link to the same commit at the bottom of the issue.

![image](https://user-images.githubusercontent.com/831974/42980155-1351e37a-8ba4-11e8-8d94-b41b5dbd76a5.png)

Setting `color: inherit` on the `a` fixes it.

```html
<div class="TableObject-item">
        <a href="#event-1743889120"><div class="State State--red">
[...]

PR #703, opened 2018-07-20
issues: better fix for #702

the-j0k3r:
See comment https://github.com/StylishThemes/GitHub-Dark/commit/8674b5577b79cb31abfe92cbde400b39710c8b02#commitcomment-29775719

After  8674b55 _fix_
capture

With this PR
capture

Original
capture

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -3081,7 +3081,8 @@
   .manage-member-meta-item .btn-link, .note, .delete-topic-button,
   .topic-tag-action .add-topic-button, .topic-tag-action .remove-topic-button,
   .text-gray-light, a.text-gray-light, .js-notice .octicon-x,
-  a.footer-octicon:hover, .table-of-contents .toc-diff-stats .octicon {
+  a.footer-octicon:hover, .table-of-contents .toc-diff-stats .octicon,
+  .issues-reset-query .octicon-x {
     color: #7b7b7b !important;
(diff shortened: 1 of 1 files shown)


the-j0k3r:
@xt0rted

the-j0k3r:
Np. Yw.

silverwind:
Actually, there's an even better fix. If you inspect the GitHub styles, you'll see they apply `.octicon, svg.octicon {fill: currentColor}`, which make the color inherit. So we just need to make sure that original style is not modified, I'll see if I can clean this up.

silverwind:
Turns out the matter wasn't as simple as I thought: I made some adjustments in https://github.com/StylishThemes/GitHub-Dark/commit/6c04b6652ef3b0e61fa0f5151b24bab267dc1517, its now much closer to GitHub's styling:

screenshot 2018-07-20 at 18 41 31

the-j0k3r:
I cant make heads or tails of it, some dialogs/modals close octicons on hover are white and some are blue after all, I think someone at GH maybe smoking too much green tea.

silverwind:
Yes, it's not always consistent. My recommendation for you would to to check first if you can fix a issue by removing a selector, before adding one. I'd say that around 50% of issues like that can usually be fixed via removal. This case was a bit different, not a good example.

Issue #665, opened 2018-07-02
review comment target brown background?


the-j0k3r:
https://github.com/StylishThemes/GitHub-Dark/blob/1a58bd914cbeae57d15c0941a5f3d1b8494f2fe2/github-dark.css#L3176

I keep seeing this brown color between the reviews comments,, It really feels somewhat out of place and default style has no such distinction.

Would `review-comment:target` be more suitable in the blue category? What do you think? Do you think blue better than *this* brown for this? IF there must be such distinction, that is.

the-j0k3r:
That came from this https://github.com/StylishThemes/GitHub-Dark/commit/32b1247fc61b48d92d0573450610c60ed7badc7f

Issue #603, opened 2018-04-04
Background color conflict with refined github in review comments


targos:
* **Browser**: Chrome 65
* **Operating System**: Windows 10
* **Screenshot**:

![image](https://user-images.githubusercontent.com/2352663/38305382-006d2f02-380d-11e8-95b6-78607ad2ffcc.png)

Link: https://github.com/nodejs/node/pull/19605#discussion_r179105528

* **HTML of the section where the issue occurs**:

````html
<div id="discussion_r179105528" class="review-comment previewable-edit js-comment js-task-list-container unminimized-comment js-reorderable-task-lists reorderable-task-lists timeline-item-highlight" data-body-version="d5aed4c6ac7381f3311ac21b66dd5400">
    </div>
````

![image](https://user-images.githubusercontent.com/2352663/38305562-8e6040d8-380d-11e8-95b4-454432a3d86e.png)

Rules added in refined github: https://github.com/sindresorhus/refined-github/pull/1182

Mottie:
Hi @targos!

Thanks for reporting this problem. This'll get fixed in the next update.

PR #707, opened 2018-07-21
re-tweak #603-#665 .review-comment:target

the-j0k3r:
ref: comments on #666 

@silverwind @xt0rted

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1973,6 +1973,10 @@
     background: #181818 !important;
     border-color: #343434 !important;
   }
+  .review-comment:target {
+    background-color: #181818 !important;
+    box-shadow: inset 0 0 0 1px /*[[base-color]]*/ #4183c4 !important;
+  }
   #readme .plain, .readme .markdown-body, .readme .plain, .pull-request-composer,
   .drag-and-drop, .octofication .broadcast-icon-mask, .file .image,
   .merge-pr-more-commits, .coupons .coupon-form-body, .manage-repo-access-group,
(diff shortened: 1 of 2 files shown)


xt0rted:
The other style rules in my comment were to fix some other colors that `refined-github` adds in.

This is what it looks like with just the border color applied.

image

This is what it looks like with the border, no background, and the extra buttons fixed.

image

the-j0k3r:
> The other style rules in my comment were to fix some other colors that refined-github adds in.

Eh forgot I dont see that since I dont use RGH :/ gimme a sec

the-j0k3r:
Is that OK or do you also strictly need a background in `.review-comment:target`?

xt0rted:
`.review-comment:target` needs a background color set

xt0rted:
An alternative to the border, which does resize the element since it normally doesn't have one, is to use `box-shadow` like so

```css
box-shadow: inset 0 0 0 1px #884;
```

Mottie:
I might be better to use the default blue `#4183c4` instead of yellow/orange since that makes it feel like a warning.

the-j0k3r:
> I might be better to use the default blue #4183c4 instead of yellow/orange since that makes it feel like a warning.

How about base color so everyone is happy? and go for the box-shadow instead?

xt0rted:
re: blue https://github.com/StylishThemes/GitHub-Dark/pull/666#issuecomment-402490825

Here's it with the background set as well, but the color muted a bit. The background is `rgba(136, 136, 68, .1)` and box-shadow is `rgba(136, 136, 68, .6)`

image

the-j0k3r:
That looks similar to the :poop: brown as before tbh :)

the-j0k3r:
Im off for tonight,whatever decisions that are made, Ill be able to update tomorrow.

Nite guys.

silverwind:
Thanks. In the future, please don't include unrelated commits in PRs :)

the-j0k3r:
OK. np.

PR #744, opened 2018-09-27
Add styling for scrollbar corner of code editor

HerrHulaHoop:
Previously, the scrollbar corner was not addressed by the CSS and appeared pure white. No the colour matches the thumbs.

Here's a comparison of the changes:
Before
before

After
after

Do let me know if anything else is required!

### DIFF
--- a/github-dark.user.css
+++ b/github-dark.user.css
@@ -3901,7 +3901,7 @@
     max-height: 10px !important;
     background: #1d1d1d !important;
   }
-  ::-webkit-scrollbar-track, ::-webkit-scrollbar-corner {
+  ::-webkit-scrollbar-track, ::-webkit-scrollbar-corner, .CodeMirror-scrollbar-filler {
     background: #1d1d1d !important;
   }
   ::-webkit-scrollbar-thumb {


silverwind:
Might even make it a shade darker or have it match the scrollbar backdrop.

HerrHulaHoop:
Well, matching it to the scrollbar backdrop gives us this:
match

and when the bars touch it, it will look like this:
fadebars

I actually quite like it with the square matching the backdrop. Let me know if you're okay and I'll make the change. Or if you want me to use another colour, you can suggest one too.

silverwind:
Matching color would be preferred, it looks better aestetically.

HerrHulaHoop:
Ok, made the necessary changes.

~~Hopefully you guys didn't see me suffer horribly with git squash~~

Mottie:
@HerrHulaHoop Please modify the `github-dark.css` file in the future. The `github-dark.user.css` is built from the `github-dark.css` file so any changes would be lost.

Also, because the change was made to a line containing `::-webkit-`, the entry would be completely ignored in Firefox. So I moved it to a separate entry.

silverwind:
Good catch, should  have looked  closer. The webkit thing is particularily important. I wonder if we can lint for it.

PR #747, opened 2018-10-02
Darken text from e2e2e2 to c0c0c0

seansfkelley:
This matches most of the other text present in the theme.

Before:

image

After:

image

Looks subtle here, makes a big difference when you have a wall of it like on the issues page there.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2861,7 +2861,7 @@
   .select-menu-item[aria-selected="true"],
   .select-menu-item[aria-selected="true"].selected, .btn-danger:hover,
   .btn-danger:active, .btn-danger.selected, [open] > .btn-danger {
-    color: #e2e2e2 !important;
+    color: #c0c0c0 !important;
   }
   pre, body, a.social-count, span.social-count, #languages a.bar, .lineoption p,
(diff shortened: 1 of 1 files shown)


silverwind:
Sorry, but I can't approve this. If you change a single color group you reduce differentian with other colors and in some locations meaning conveyed through color will get. I think the only change I'd accept is a slight darkening of all `color` rules by maybe around 6 hex numbers, but it needs to be done universally (also in `generate.js`).

seansfkelley:
What about if all the colors were darkened by a bit and `#e2e2e2` was darkened to something darker-but-not-as-dark as the other text? It's pretty close to white at the moment.

Mottie:
@seansfkelley If you only want to darken the text of the issue list, then move the specific selector instead of changing the color definition for everything.

seansfkelley:
@Mottie there's a lot of text of this color, including headers in the readme, usernames in some contexts, numbers along the repo summary page... it's splattered all over the place, which is why I find it hard on the eyes.

silverwind:
I went ahead and darkened text by 6 hex digits. I consider this the maximum, otherwise text will look too dull.

My initial feeling is that it's already too dull after this change but I'll use it a few more days, then we can decide wether to keep it.

PR #749, opened 2018-10-07
Add css to properly colorize directory icon

phallstrom:
Noticed the directory icons weren't being properly styled, due to a
missing style. They are now styled to match file and symlink icons.

See the before and after shots.

before
after

It's not clear to me if I've patched this in the proper way based on some other commits so if I need to adjust this somewhere else please let me know. Thanks.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -3036,7 +3036,7 @@
   .dashboard-notice .dismiss:hover, .text-gray, td.icon .octicon-file-text,
   td.icon .octicon-file-symlink-file, .UnderlineNav-item, .project-header-link,
   button.project-header-link, td.icon, .project-pane-close, .facebox-close,
-  .tree-browser-result mark {
+  .tree-browser-result mark, td.icon .octicon-file-directory {
     color: #949494 !important;
   }
   .vcard-detail .octicon, .member-badge .octicon,


github-dark.css
```diff
@@ -3033,8 +3033,9 @@
...
   .org-header-meta .meta-item .meta-link, .underline-nav-item,
   .boxed-group-inner, .lead, .form-checkbox .note, .bot-identifier,
-  .dashboard-notice .dismiss:hover, .text-gray, td.icon .octicon-file-text,
-  td.icon .octicon-file-symlink-file, .UnderlineNav-item, .project-header-link,
+  .dashboard-notice .dismiss:hover, .text-gray,
+  td.icon .octicon-file-text, td.icon .octicon-file-symlink-file, td.icon .octicon-file-directory,
+  .UnderlineNav-item, .project-header-link,
   button.project-header-link, td.icon, .project-pane-close, .facebox-close,
   .tree-browser-result mark {
     color: #949494 !important;
```
Mottie: Please add the above change to the end of the selector list (i.e. move it next to the opening `{`) to make it easier to see the change.

phallstrom: Done

Mottie:
Thanks!

silverwind:
Hmm I don't see this issue, also we had directories on the base color:

screenshot 2018-10-16 at 17 28 39

The icon is styled via `table.files .octicon-file-directory` for me. @Mottie were you been able to reproduce?

Mottie:
The directory icons match the file icon colors for me since I merged this PR.

silverwind:
Weird, I'm not seeing this in Firefox.

silverwind:
Also checked in Chrome and it looks like this if I revert this patch:

screenshot 2018-10-16 at 17 47 07

@phallstrom what browser were you using and were you up to date?

silverwind:
Testing repo with symlinks here: https://github.com/silverwind/symlink-test

Mottie:
I'm seeing this:

Chrome


Firefox

silverwind:
Okay I'll revert this patch because I like colored directories. I think OP might have been running on an old version or a weird browser. We can still fix it if it is an actual issue.

phallstrom:
@silverwind Running the latest Safari on Mojave.

silverwind:
@phallstrom which userstyle extension? Seems I can no longer install Stylish there:

screenshot 2018-10-16 at 18 44 12

phallstrom:
@silverwind http://cascadea.app

xt0rted:
You can replace the default icons entirely with https://github.com/xxhomey19/github-file-icon. Doesn't effect the GitHub Dark styles at all either.

silverwind:
@xt0rted nice extention, I'll use it 👍

phallstrom:
@silverwind Looks like you already merged this, but this is what I see without my patch.

screen shot 2018-10-17 at 1 36 20 pm

silverwind:
I see the GitHub Dark selector matching in your screenshot, but it appears to not win against the site style. Could you scroll just a tad bit down until our color statement? I suspect your userstyle engine might ignore/drop our `!important` declaration.

phallstrom:
@silverwind this is what I see (with some gaps that don't have the color attribute so i left off)

screen shot 2018-10-18 at 11 01 09 am
screen shot 2018-10-18 at 11 01 33 am

silverwind:
Interesting, your rule is missing the fallback color. It should be

````css
color: /*[[base-color]]*/ #4183c4 !important;
````

What method of installation did you use?

phallstrom:
I think I just went to stylish and followed the urls. The one it's using is this.  Now I'm wondering if I've messed that up somehow.

https://github.com/StylishThemes/GitHub-Dark/raw/master/github-dark.user.css

Mottie:
Stylish doesn't know how to deal with [UserCSS](https://github.com/openstyles/stylus/wiki/UserCSS) files. You'll need to install [Stylus](https://add0n.com/stylus.html) to get that file to work properly.

phallstrom:
Ah, so perhaps I should bring this up with Cascadeapp.  It's meant to work with them, but perhaps not completely supported.

silverwind:
Proper UserCSS support is needed in your engine for customization, but you should be able to use the development style which has the fallback colors:

https://raw.githubusercontent.com/StylishThemes/GitHub-Dark/master/github-dark.css

phallstrom:
Much better! Thanks

PR #756, opened 2018-10-16
Fix zenhub reports

Kay-Zee:
Realized that I missed some of the reports, this fixes the Cumulative Flow and Release Report pages. Also fix the invite team modal.


Mottie:
Awesome! Thanks again!

PR #762, opened 2018-10-28
Fix Zenbub nav tab icon colors

olets:
Currently the Zenhub tab's icon is dim when the Zenhub tab is selected, and bright otherwise (ignore my custom body background color in the screenshots):

screen shot 2018-10-28 at 15 07 49
screen shot 2018-10-28 at 15 07 59

With this PR, the Zenhub tab's icon is bright icon when the Zenhub tab is selected, and dim otherwise:

screen shot 2018-10-28 at 15 11 20
screen shot 2018-10-28 at 15 10 29

The new colors as close to the "correct" colors as is possible using `filter` in Chrome 70, even with fractional `brightness`: `#b9babc` for bright, `#757678` for dim (compared to the native tab icons' `#bababa` and `#757575`)

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -4265,9 +4265,14 @@
     height: 14px !important;
   }
   /* invert image */
-  .zh-icon-circled-bolt, .zh-icon-bolt, .zh-octicon-x,
-  a.zh-topbar-item.selected .zh-icon-zenhub {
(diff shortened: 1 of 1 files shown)


Mottie:
Awesome, thanks @olets!

PR #772, opened 2018-11-18
Make file headers sticky everywhere (fixes #771)

### LINKED ISSUE #771: Make file headers sticky on commit and compare view just like on PR file list
The sticky file headers requested for the PR file list in #531 and implemented in 5b91c21f18 are very useful.

Please make the file headers sticky on [commit](../commit/fdbd89afd6) and [compare](../compare/master@{5day}...master) pages too, so you always see what file you are looking at.

ght:
This makes the file headers on commit and compare pages sticky too.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1359,11 +1359,14 @@
     top: 8px !important;
   }
   @supports (position: sticky) or (position: -webkit-sticky) {
-    .pull-request-tab-content .diff-view .file-header {
+    .diff-view .file-header {
       position: sticky !important;
       position: -webkit-sticky !important;
-      top: 40px !important;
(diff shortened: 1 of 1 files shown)


Mottie:
Oh sorry, I pushed some changes that are conflicting with this PR. Would you please update it... override my changes as needed.

the-j0k3r:
@Mottie 65f3e82b1c8e9c3d09ee7aa73840577ee0c38021 closes this so we cant add it again unless its its own style, right?

the-j0k3r:
@ght This PR should be reopened against the new sticky headers style as soon as its decided where this code is going to live. So theres hope yet for the fans.

PR #794, opened 2018-12-08
[wip] Testing out simpler input styling

xt0rted:
This is in response to https://github.com/StylishThemes/GitHub-Dark/commit/13765606e236ac4051f5d73d4b1b0b8e804aefab#commitcomment-31604759

I've been testing this out on github.com, gist.github.com, and with Octotree enabled. So far things seem good but I'm leaving this open to test it some more (I think there's more selectors that can be removed) and to collect feedback on it.

If we can get Octotree to use the `form-control` class on their inputs then we should be able to remove the selectors for their inputs too.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1276,34 +1276,21 @@
     outline-color: /*[[base-color]]*/ #4f8cc9 !important;
   }
   input.focus[type="text"], #adv_code_search .focus.search-page-label,
-  /* below selector has a ton of weight. Better to override it with
-   * class selectors that unset box-shadow and rules here. Testable
-   * on the repo tag editor */
-  input[type="text"]:focus:not(.js-color-editor-input):not(.tree-finder-input):not(.js-site-search-focus):not(.tag-input-inner),
   .focused .drag-and-drop, #adv_code_search .search-page-label:focus,
-  input.focus[type="password"], input.form-control[type="password"]:focus,
-  input.focus[type="email"], input.form-control[type="email"]:focus,
-  input.focus[type="number"], input.form-control[type="number"]:focus,
-  input.focus[type="tel"], input[type="tel"]:focus, input.focus[type="url"],
-  input.form-control[type="url"]:focus, textarea.focus, textarea:focus,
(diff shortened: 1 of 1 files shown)


github-dark.css
```diff
@@ -1276,35 +1276,22 @@
...
     border-color: /*[[base-color]]*/ #4f8cc9 !important;
     box-shadow: inset 0 1px 2px rgba(0, 0, 0, .075), 0 0 2px /*[[base-color]]*/ #4f8cc9 !important;
     outline-color: /*[[base-color]]*/ #4f8cc9 !important;
   }
   .dragover textarea, .dragover .drag-and-drop {
     box-shadow: 0 0 1px 1px /*[[base-color]]*/ #4f8cc9 !important;
   }
-  /* inputs */
-  select, input:not(.btn):not(.btn-link):not(.btn-danger), textarea {
-    -moz-appearance: none !important;
```
silverwind: Looks like this is still neccesary to style `input[type=checkbox]` on Firefox. First checkbox below does not have it:

screenshot 2018-12-10 at 19 40 22

xt0rted: Is that on the PR page? If so I'm running FF 65.0b2 (64-bit) and not seeing the issue.

Actually, with GHD disabled I don't see checkboxes on that page at all.

xt0rted: Interesting, in Chrome those checkboxes show, but in Edge & FF they don't. In FF there's a space for them, and in Edge there isn't.

silverwind: That's from the PR page, yes. This may be related to me running Firefox 57. Oh, and the same goes for `input[type="radio"]`.

silverwind: Actually, it appears Mozilla now supports our existing `-webkit-appearance`, but not in version 57. Can you just add `-moz-appearance` in the place we have `-webkit-appearance` for checkboxes/radios.

xt0rted: Looks like FF was caching part of GHD even when it was disabled, that's why my checkboxes weren't showing up.

PR #807, opened 2018-12-27
Remove unneeded width style from user timeline header

xt0rted:
PRing this to get some feedback. From what I can tell removing this rule doesn't effect the user profile timeline. Maybe there's something I'm missing but it seems like it's safe to remove now.


silverwind:
Should be fine. If I remember correctly, this was a workaround to a GitHub issue where this line overflowed into the year list on the right.

Issue #68, opened 2013-11-20
Language statistics colors


silverwind:
I'd like to recolor the language-colors to more flashy colors so they stand out more on dark backgrounds. Unfortunately, GitHub applies them through inline styles, so we need to create a list of colors in use first. Here's what I have so far from my projects:

````css
background-color:#f15501; /* javascript */
background-color:#1f085e; /* css - hard to read */
background-color:#fa1fa1; /* lua */
background-color:#5a25a2; /* c# */
````

Mottie:
Language colors have some minor changes, and some definitions are missing:

Language | Color
------------ | -------------
AGS Script | "#B9D9FF"
AMPL | "#00008B"
ANTLR | "#9DC3FF"
APL | "#8a0707"
ASP | "#6a40fd"
ATS | "#1ac620"
ActionScript | "#e3491a"
Ada | "#02f88c"
Agda | "#467C91"
Alloy | "#cc5c24"
Arc | "#ca2afe"
Arduino | "#bd79d1"
AspectJ | "#1957b0"
Assembly | "#6E4C13"
AutoHotkey | "#6594b9"
AutoIt | "#36699B"
BlitzMax | "#cd6400"
Boo | "#d4bec1"
C | "#555"
C# | "#178600"
C++ | "#f34b7d"
CSS | "#563d7c"
Chapel | "#8dc63f"
Cirru | "#aaaaff"
Clean | "#3a81ad"
Clojure | "#db5855"
CoffeeScript | "#244776"
ColdFusion | "#ed2cd6"
ColdFusion CFC | "#ed2cd6"
Common Lisp | "#3fb68b"
Component Pascal | "#b0ce4e"
D | "#fcd46d"
DM | "#075ff1"
Dart | "#98BAD6"
Dogescript | "#cca760"
Dylan | "#3ebc27"
E | "#ccce35"
ECL | "#8a1267"
Eagle | "#3994bc"
Eiffel | "#946d57"
Elixir | "#6e4a7e"
Elm | "#60B5CC"
Emacs Lisp | "#c065db"
EmberScript | "#f64e3e"
Erlang | "#0faf8d"
F# | "#b845fc"
FLUX | "#33CCFF"
FORTRAN | "#4d41b1"
Factor | "#636746"
Fancy | "#7b9db4"
Fantom | "#dbded5"
Forth | "#341708"
Frege | "#00cafe"
Game Maker Language | "#8ad353"
Glyph | "#e4cc98"
[...]

PR #810, opened 2018-12-30
changing language-color doesnt work as well as intended

dym-sh:
not to mention _brighten_ it to the point there a swift repository would have a javascript language-color

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -2531,10 +2531,6 @@
   .contribution-first-repository {
     opacity: .999 !important;
   }
-  /* brighten language colors */
-  .language-color {
-    filter: brightness(150%) !important;
-  }
   /* soft badges */
   .box-title-count {
     background-color: #484848 !important;


the-j0k3r:
@xt0rted thoughts on ways to make this better?

xt0rted:
@dym-sh can you give us a link to a page where this is an issue?

dym-sh:
@xt0rted here is a color-case-study:

JS in the listings:
screenshot_2019-01-03 build software better together

Swift in the listings:
screenshot_2019-01-03 build software better together 1

Swift in the repo:
screenshot_2019-01-03 vsouza awesome-ios

@the-j0k3r 
> ways to make this better
i think the color-correction was simply unnecessary change since it brought nothing but confusion, hence plain deletion of lines responsible for that.

also, to _brighten_ colors on a dark theme seems like a counter-intuitive move.

xt0rted:
After browsing some repos with the filter off I personally prefer it that way, especially because of this JS/Swift similarity. But also because these colors are used in other spots and keeping the filter (or changing the colors to something else) would mean adjust those spots too and it seems easier to leave them as-is.

I'm giving removing this a 👍

@Mottie & @silverwind what are your thoughts on this one?

For reference this change was originally made in 81bbd03038b644b938745eecf039bc11762d5a0d for #68

xt0rted:
Looks like PowerShell `#012456` is harder to see without the filter. We should be able to adjust that in all spots with:
```css
/* GitHub doesn't put a space between the : and # so we can't either */
[style*="background-color:#012456;"] {
  filter: brightness(150%) !important;
}
```

Both `*=` and `$=` work here, not sure which is better for this type of thing though.

dym-sh:
i spotted several similar dark-blue colors throughout the https://github.com/ozh/github-colors

`#00004c` https://github.com/trending?l=Terra
`#000080` https://github.com/trending?l=Lua

`#04133b` https://github.com/trending?l=WebAssembly
`#012456` https://github.com/trending?l=PowerShell

but brighten the first pair will make them look like the second one, and the second one will look like `#244776` https://github.com/trending?l=CoffeeScript

the-j0k3r:
Would it be possible to override all the available language colors to something that would work better globally with GitHub Dark and make colors more visible?

xt0rted:
@the-j0k3r see 81bbd03038b644b938745eecf039bc11762d5a0d, that's how this was originally handled. The filter was meant to simplify the rules and trim the size of the style but it doesn't work well in all situations. We could go back to a hand picked list of colors but I'm not writing all that css 😛

the-j0k3r:
> We could go back to a hand picked list of colors but I'm not writing all that css 😛

Thats the easy part, a copy and pasta and away we go... Ill have a look and push a PR, thx for digging up the original commit. (assuming that css still works.)

silverwind:
The brightening is primarly to make the darker colors more visible and given that these color changes have been in the repo for more than 5 years, I'd wager a few people have gotten used to them by now (I certainly have).

I think the real bug here is that those language dots are not styled in the same color everywhere which we should fix. Maybe we could also reduce the brightness filter slightly to 125%, if it helps the darker colors, that is.

xt0rted:
Here's some side by side examples of the issue the filter is causing. This is made even worse for anyone using a program like [f.lux](https://justgetflux.com/) or Windows Nightlight (which I am). There's been tweaks to GitHub Desktop's dark theme for this very reason too.

I know removing the filter entirely starts to break some of the darker colors but over all I found it better without it than with it which is why I have a custom style that removes the filter now.

**With the filter** (JS, Swift)

image

**Without the filter**

image


**With the filter** (JS, Shell, Swift)

image

**Without the filter**

image

silverwind:
> There's been tweaks to GitHub Desktop's dark theme for this very reason too.

Is that code public somewhere?

> Here's some side by side examples

Agree, this is not ideal. Does reducing the filter intensity help? Maybe 125%.

the-j0k3r:
See https://github.com/StylishThemes/GitHub-Dark/commit/ebc035e1d88789069912d566495167a1718286c7 interim solution, should be out next release though I have a few things to get in before then.

xt0rted:
@silverwind The desktop app has "experimental" dark theme support which was added in https://github.com/desktop/desktop/pull/4849 and has been improved upon ever since. One of the PRs I know of that adjusted the colors is https://github.com/desktop/desktop/pull/5094 which was a direct response to complaints in https://github.com/desktop/desktop/issues/5027. It's been awhile since I followed this this repo though so I'm not sure if there's any other issues/PRs that better address this.

the-j0k3r:
@xt0rted desktop/desktop#5094 +  desktop/desktop#5027 are about diff colors not language colors, am I missing something here?

the-j0k3r:
If we make this optional, (which I dont mind personally) the actual defaults suck just as bad,

The alternative is like @silverwind said and style just the problematic ones that are already too close.

I dont mind doing that, but at this moment I would ask for any collaborator or person willing to help out in the actual PR.

or feel free to change this PR and make it optional.

silverwind:
I think the 125% is a good compromise.

PR #815, opened 2019-01-04
fix: custom language colors closes #810

the-j0k3r:
note: reintroduces https://github.com/StylishThemes/GitHub-Dark/commit/e729acecb004ece241e0b5cc0ea72283438898e3 by @silverwind 

thx to @xt0rted for having dug up this treasure and sharing. =)

Close #810 (see for discussion)


xt0rted:
just got confirmation that the colors for the languages come from this file https://github.com/github/linguist/blob/e761f9b013e5b61161481fcb898b59721ee40e3d/lib/linguist/languages.yml

the-j0k3r:
@dym-sh last commit accounted for the blue colors you mentioned in https://github.com/StylishThemes/GitHub-Dark/pull/810#issuecomment-451342586

Generally this will work better, I also think the colors need tweaking here and there in this PR, perhaps a job for the long run...

Ill wait for feedback.

>This is rebased on master as of 1.20.96 + userstyles link removal.

dym-sh:
the only thing missing is settings-switch for turning alternative language-colors on/off

the-j0k3r:
> the only thing missing is settings-switch for turning alternative language-colors on/off

Thats a good idea, I can push this in that direction, the question is on by default or off by default.

the-j0k3r:
Ok after banging my head a bit and tuning (so excuse the force-push noise), this is now optional

optionalclc

Please test usercss via direct install by [clicking here](https://raw.githubusercontent.com/the-j0k3r/GitHub-Dark/custom-language-colors/github-dark.user.css)
>Note UserCSS wont be part of merge(if this is merged) its just for testing.

dym-sh:
tested in chrome, all perfect 👍

the-j0k3r:
> tested in chrome, all perfect +1

The option was working OK, but a couple of colors had broken selectors, Ive just rebased on master and fixed the issue + cleanup.

Now we wait for feedback from the masters of the universe before delivering this baby.

silverwind:
Not sure I like this approach, it will require maintenance and I'd generally prefer a more automatic approach instead. How about we just modify the few dark ones, or go with a slightly less intensive `brightness` filter?

the-j0k3r:
Not sure, Maintenance wise its minimal, its not like something that changes often, that said, Im ok with whatever you decide.

silverwind:
I'd say go with 125% brightness and make sure those color dots are styles too.

PR #821, opened 2019-01-07
Add generate rule for text-blue-mktg

xt0rted:
PRing for some feedback. This fixes the blue links on the [new pricing page](https://github.com/pricing), but seems to break some of them elsewhere such as on https://github.com/features. The issue is they're using `box-shadow` on `:hover` which is being overridden in another rule. Any ideas on what we should do to handle these?

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -824,6 +824,12 @@
   .illflow-blue .illflow-item .illflow-item-heading, .MarketingBody-lead a:hover {
     color: /*[[base-color]]*/ #4f8cc9 !important;
   }
+  /* auto-generated rule for "color: #1277eb" */
+  .text-blue-mktg, .link-mktg, .btn-outline-mktg,
+  .casestudy-link:hover .casestudy-title, .MarketingBody blockquote,
+  .MarketingBody-lead a {
+    color: /*[[base-color]]*/ #4f8cc9 !important;
+  }
   /* auto-generated rule for "color: #586069" */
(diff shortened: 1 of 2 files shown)


the-j0k3r:
I cant really tell where it breaks, perhaps its too nuanced for my eyes.

the-j0k3r:
This is somewhat related to #828

silverwind:
If one generated style is interfering with another generated one of the same specificity, it maybe possible to adjust the order in the `mappings` object. Otherwise I guess you'll need to add a manual fix using a non-generated rule.

the-j0k3r:
Any news on this, it would fix logout `.link-mktg` color as well.

silverwind:
I'd say go ahead with this and also remove the `a { box-shadow: none }` rule we have. Also, I'd add another generator rule to replace the color in these box-shadow underlines.

the-j0k3r:
This should also fix https://help.github.com/en/categories/github-pages-basics `.text-blue-mktg`

xt0rted:
Do you still need me to look into the other rules for this? I can do that later today if needed.

silverwind:
Yes, if you like. Thought I'm quite happy with how those pages look already.

xt0rted:
Looks like you already removed the box shadow 👍

the-j0k3r:
Speaking of box shadows, here is a weird one.

URL: https://github.com/about

padding-top

On hover the blue is GH default, can this be fixed in autogenerated?

Issue #651, opened 2018-06-19
really blinding blue button


the-j0k3r:
* **Browser**: FF 66.0.2 blablabla
* **Operating System**: winblows 10 bild the build 1709
* **Screenshot**:

![capture](https://user-images.githubusercontent.com/31389848/41628149-2ae03f78-741b-11e8-8cbf-f7579088b06b.PNG)

* **HTML of the section where the issue occurs**:

````html
<form action="/settings/security/delegated-account-recovery" accept-charset="UTF-8" method="get"><input name="utf8" value="✓" type="hidden">
            <input name="provider_origin" value="https://www.facebook.com" type="hidden">
            <input src="https://assets-cdn.github.com/images/modules/dar/store-token-with-facebook.png" width="236" type="image" height="34">
</form>
````

the-j0k3r:
This is probably not a bug, but WHY would anyone trust facebook with security?

Mottie:
Crazy that they're using an input as an image... not much we can do about it. Here's what it looks like inverted & hue rotated:

![](https://user-images.githubusercontent.com/136959/41628324-9aff0e06-73e9-11e8-9447-7e3e56e85f9f.png)

the-j0k3r:
lol invert and hue-rotate are not the only tricks in the bag.

try: filter: grayscale/contrats and brightness if you need it.

Mottie:
And just because... here it is without a width restriction LOL

![](https://assets-cdn.github.com/images/modules/dar/store-token-with-facebook.png)

the-j0k3r:
![capture](https://user-images.githubusercontent.com/31389848/41628462-8865b7e4-741c-11e8-9e53-aa4b0c9f001f.PNG)


```filter: grayscale(100%) contrast(80%) !important```

Mottie:
Reusing existing selectors...

the-j0k3r:
neat, grayscale 30% also good :

the-j0k3r:
brightness 75% is like 25% more wtf. :) hahaha good think I never go there to trust facebook with my security.

PR #825, opened 2019-01-10
Changed private repository lock symbol color

Otman404:


github-dark.css
```diff
@@ -1659,6 +1659,10 @@
   .private .mini-repo-list-item {
     background: #221 !important;
   }
+  /*Private repo Lock color*/
+  .private a svg {
+		color: #ffda79 !important;
```
the-j0k3r: I would use this selector to a existing yellow rule

e.g.

https://github.com/StylishThemes/GitHub-Dark/blob/5b619a13ff24012899dc66a85d8793268d43d83d/github-dark.css#L3431-L3433

xt0rted: I think this color is too bright for here. I'm looking at using `#651` which is part of the generator and used in a few other spots. Just checking to see what else this may effect. 

image image image

What's odd is they aren't styling this lock anywhere else (yet) and over the last year or so removed this color from the `private` label on repos.

image
image

https://github.com/StylishThemes/GitHub-Dark/blob/5b619a13ff24012899dc66a85d8793268d43d83d/github-dark.css#L3388-L3397

xt0rted: So yea, the gold/yellow used on this lock is also used for PR check statuses. The red circle below on this PR can have pending/queued/etc. statuses which we're currently not handling (looks like we're not handling any of them actually). I'm working on an update now that should fix all of these.

Default color

image

Using `#cb4`

image

Using `#651`

image

xt0rted:
I just pushed a new version (v1.20.97) which adjusts the colors of:

- The build check status donut chart
- The Code Frequency graphs under the Insights tab
- The percentages on the activity graph of your profile
- The lock icon on the new dashboard

Give it a try and let me know what you think.

Otman404:
@the-j0k3r  Sorry i forgot to include screenshots, and yeah that color looks slighly better.
@xt0rted  Thanks for the Update :)

the-j0k3r:
Superseded by #827 has this addition in correct place.

PR #827, opened 2019-01-10
Adjusting  lock and repo suggestion name color

Otman404:
Adjusting lock color when creating a private repo, and repo suggestion name color to be more visible ( yeah nobody cares about suggestion names but still :p )
Before : 
beforee

After : 
afterr

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -1674,10 +1674,6 @@
     background: #222 !important;
     color: #777 !important;
   }
-  /*Repo name Suggesiton color*/
-  .reponame-suggestion {
-    color: #539c2a !important;
-  } 
   .private .mini-repo-list-item .repo-icon {
     color: #777 !important;
   }
@@ -3502,7 +3498,8 @@
   .signed-commit-badge.verified:hover, .Label--outline-green {
     border-color: #6cc644 !important;
   }
-  .signed-commit-verified-label, .signed-commit-badge.verified {
+  .signed-commit-verified-label, .signed-commit-badge.verified,
+  .reponame-suggestion {
     color: #55a532 !important;
   }
   /* Green text */
--- a/github-dark.css
+++ b/github-dark.css
@@ -3449,7 +3449,8 @@
   .text-pending, .text-renamed, .bg-pending {
     color: #cb4 !important;
   }
-  .page-new-repo .octicon-lock {
+  .page-new-repo .octicon-lock,
+  .private a svg {
     color: #651 !important;
   }
   /* yellow border (travis) */
--- a/github-dark.css
+++ b/github-dark.css
@@ -1674,14 +1674,9 @@
     background: #222 !important;
     color: #777 !important;
   }
-  /*Create private repo Lock color*/
-  .page-new-repo .octicon-lock {
-    color: #cb4 !important;
-  }
   /*Repo name Suggesiton color*/
   .reponame-suggestion {
     color: #539c2a !important;
-    cursor: pointer;
   } 
   .private .mini-repo-list-item .repo-icon {
     color: #777 !important;
@@ -3473,6 +3468,10 @@
   .Label--outline.bg-yellow-light,
   .review-comment .pending-batched-suggestion-label {
     background: #302808 !important;
+  }
+  /* Yellow private repo lock */
+  .page-new-repo .octicon-lock {
+    color: #cb4 !important;
   }
   /* green */
   /* labels */
--- a/github-dark.css
+++ b/github-dark.css
@@ -3446,8 +3446,7 @@
   }
   /* yellow text */
   /* .bg-pending might be a GitHub bug as it sets the fg */
-  .text-pending, .text-renamed, .bg-pending, .page-new-repo .octicon-lock,
-  .private a svg {
+  .text-pending, .text-renamed, .bg-pending, .page-new-repo .octicon-lock {
     color: #cb4 !important;
   }
   /* yellow border (travis) */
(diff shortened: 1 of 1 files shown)


github-dark.css
```diff
@@ -1674,6 +1674,15 @@
     background: #222 !important;
     color: #777 !important;
   }
+  /*Create private repo Lock color*/
+  .page-new-repo .octicon-lock {
```
xt0rted: Can you move this selector to the end of https://github.com/StylishThemes/GitHub-Dark/blob/6c61d52cc041c5b4693164754c6ab6ac716d3ff2/github-dark.css#L3449

github-dark.css
```diff
@@ -1674,6 +1674,15 @@
...
+    color: #cb4 !important;
+  }
+  /*Repo name Suggesiton color*/
+  .reponame-suggestion {
+    color: #539c2a !important;
+    cursor: pointer;
```
xt0rted: This isn't needed.

```suggestion
```

github-dark.css
```diff
@@ -1674,6 +1674,15 @@
...
+  }
```
xt0rted:
```suggestion
  }
```

Otman404:
c404e85c2dc074c7d2f86ce1c0b8265ac9aecf99 what do you think ?

the-j0k3r:
Again like in the other PR, I think we should use the same color for padlock icons, I think `color: 651` is better than a bright urine yellow color.

github-dark.css
```diff
@@ -3465,6 +3469,10 @@
   .review-comment .pending-batched-suggestion-label {
     background: #302808 !important;
   }
+  /* Yellow private repo lock */
+  .page-new-repo .octicon-lock {
+    color: #cb4 !important;
```
the-j0k3r:
```suggestion
    color: #651 !important;
```

the-j0k3r: I would also make the selectors for this PR and #825 part of the same rule so we dont end up with duplicate color rules with only one selector for same job.

github-dark.css
```diff
@@ -1674,6 +1674,10 @@
     background: #222 !important;
     color: #777 !important;
   }
+  /*Repo name Suggesiton color*/
+  .reponame-suggestion {
+    color: #539c2a !important;
```
the-j0k3r:
```suggestion
    color: #539c2a !important;
```

We have plenty of greens you can add this to, for instance `#55a532` so please add this selector to
https://github.com/StylishThemes/GitHub-Dark/blob/6c61d52cc041c5b4693164754c6ab6ac716d3ff2/github-dark.css#L3498-L3500

You dont need the comment either.

the-j0k3r:
@Otman404  I added a couple of commits addressing comments by myself and @xt0rted so lets see what he thinks..

Otman404:
@the-j0k3r  yes, that looks good

the-j0k3r:
@Otman404 OK I added the change from #825 into this PR, its a trivial change to add, so Ill be closing #825

the-j0k3r:
@xt0rted have a glimpse and if OK just kick in.

@Otman404 thank you for your patience. =)

xt0rted:
@silverwind the color used on these lock icons is used for the build status charts. If you go with the darker color then those don't look right which is why I used the lighter version in the generator (03c9e8cd18922fee591ff4e69b42a5bdaffebd41). Since they're using this color in another spot (the dashboard & repo popovers on the dashboard) I'm betting they're going to start using it more in other spots too.

silverwind:
Ah, alright. In that case a exception from the norm seems in order :wink:

xt0rted:
Or we could keep the lighter color on the locks too 😜

xt0rted:
`#cb4` showed up in another spot 😛

image

the-j0k3r:
@xt0rted I vote `#651` its not urine colored

xt0rted:
`#651` is too dark. Icons of this color either blend in to the background, or look like they're in a disabled state. This is made worse when using something like f.lux where the color looks even more muted at times.

The point of them using yellow/gold for these icons is to highlight their higher level of importance from the default state. `#cb4` does this much better, as does `#884` which is the yellow we're using on the pending build status box.

the-j0k3r:
pushed https://github.com/StylishThemes/GitHub-Dark/pull/827/commits/338ee06f5ddcb44de4964a90d83021d040c22b36 as @xt0rted indicated.

github-dark.css
```diff
@@ -3446,7 +3446,8 @@
   }
   /* yellow text */
   /* .bg-pending might be a GitHub bug as it sets the fg */
-  .text-pending, .text-renamed, .bg-pending {
+  .text-pending, .text-renamed, .bg-pending, .page-new-repo .octicon-lock,
+  .private a svg {
```
xt0rted: I don't think this selector is needed (it's for the lock on the dashboard right?). If so it should be covered by `.repo-private-icon` in https://github.com/StylishThemes/GitHub-Dark/blob/68bcb8d7cea5166b35035b5f77edf7500180610d/github-dark.css#L1155-L1159

the-j0k3r: See https://github.com/StylishThemes/GitHub-Dark/pull/827/commits/3aa34493eeca90e0deb927148244b9cb605b1de6 removed.

Anything else?

the-j0k3r: @xt0rted anything else?

the-j0k3r:
Do we need anything here before merging?

xt0rted:
Looks good to me

Issue #830, opened 2019-01-11
You can customize font-family, but not font-size


LordJZ:
* **Browser**: Firefox latest
* **Operating System**: Windows 10
* **Screenshot**: n/a

* **HTML of the section where the issue occurs**: any code

You can customize `font-family`, but not `font-size`. Since different fonts have different baseline sizes, I think it makes sense to make it an option as well.

For example, I use `Fira Code` as my monospace font everywhere, but with this userstyle, the default font size is too large compared to its surroundings. I did hard-code it to `11px` myself, but I think it would still make a neat addition.

I'll try to work on this myself in the next hour or so; I will send a PR if I succeed. Thanks.

PR #831, opened 2019-01-11
Add Font Size choice

LordJZ:
This is an implementation for #830.

I wonder if it is possible to specify the "github default" as default. In other words, make the default such that github's css values are used. I couldn't find a way to do this, so I used `12px` as the default.


the-j0k3r:
>  In other words, make the default such that github's css values are used. I couldn't find a way to do this, so I used `12px` as the default.

Hi thx for PR, I think its a good addition =), but I also wonder the same thing, lets wait for a master of the universe to add something on that.

silverwind:
> make the default such that github's css values are used

Hard, but not impossible. It could be done in `tools/generate.js`, but I don't think it would be worth it given how unlikely it is that GitHub will change their code font size.

Changes look good and throurgh to me, but maybe @Mottie can have a final look.

defaults.json
```diff
@@ -4,6 +4,7 @@
   "themeJP"  : "twilight",
   "color"    : "#4f8cc9",
   "font"     : "Menlo",
+  "fontSize" : "12px",
```
LordJZ:
```suggestion
  "fontSize" : "github default",
```

github-dark.user.css
```diff
@@ -172,6 +172,7 @@
   } EOT;
 }
 @advanced text font-choice "Code Font" "Menlo"
+@advanced text font-size-choice "Code Font Size" "12px"
```
LordJZ:
```suggestion
@advanced text font-size-choice "Code Font Size" "github default"
```

LordJZ:
There you go @the-j0k3r, thanks.

the-j0k3r:
> There you go @the-j0k3r, thanks.

np, Thank you =)

github-dark.user.css
Mottie: Did you try setting the font-size to `initial` or `inherit`?

silverwind:
> font-size to initial or inherit

I don't think this will work because it would be a valid rule and would potentially override other rules on the same object. I think a invalid rule is the best way to go for the default.

LordJZ:
> 
> 
> > font-size to initial or inherit
> 
> I don't think this will work because it would be a valid rule and would potentially override other rules on the same element. I think a invalid rule is the best way to go for the default.

Yep -- that's why the default is `default` which is an invalid value for `font-size`.

Mottie:
Derp!

PR #840, opened 2019-01-22
Add Overlay-Scrollbars to Readme.md

RomanKornev:
I haven't seen any reference to the split Github Overlay Scrollbars repo in the Readme. Seems like it should belong here.


Mottie:
Thanks! The split only happened a few days ago.

PR #857, opened 2019-02-23
Fix usercss generator so it supports multiple document rules

xt0rted:
This fixes the usercss generator so it supports multiple `@-moz-document` rules which is currently being used for the payment iframes, and will also be used to style some of the actions editor.

The second commit shows what was missing by not supporting this.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -165,7 +165,7 @@ module.exports = function(grunt) {
     replacement: config.tab
   }, {
     // remove default syntax themes AND closing bracket
-    pattern: /(\s+\/\* grunt build - remove start \*\/)(.*)(\s+\/\* grunt build - remove end \*\/)$/sm,
+    pattern: /\s+\/\* grunt build - remove start \*\/.*\s+\/\* grunt build - remove end \*\/$/sm,
     replacement: ""
   }, {
     // add selected theme
@@ -199,7 +199,7 @@ module.exports = function(grunt) {
     replacement: "/*[[tab-size]]*/"
   }, {
     // remove default syntax theme AND closing bracket
-    pattern: /(\s+\/\* grunt build - remove start \*\/)(.*)(\s+\/\* grunt build - remove end \*\/)$/sm,
+    pattern: /\s+\/\* grunt build - remove start \*\/.*\s+\/\* grunt build - remove end \*\/$/sm,
     replacement: ""
   }];
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -165,7 +165,7 @@ module.exports = function(grunt) {
     replacement: config.tab
   }, {
     // remove default syntax themes AND closing bracket
-    pattern: /\s+\/\* grunt build - remove to end of file(.*(\n|\r))+\}$/m,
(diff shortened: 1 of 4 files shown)


silverwind:
Please run a `npx grunt update && rm -rf node_nodules && npm i` and commit the result to hopefully fix the CI error (I bumped `eslint-config-silverwind` to support ecmaVersion 2019 and the `s` flag on regex).

Gruntfile.js
```diff
@@ -339,7 +339,7 @@ module.exports = function(grunt) {
       mozrule: {
         files: {"<%= config.buildFile %>" : "<%= config.buildFile %>"},
         options: {
-          wrapper: ["<%= config.prefix %>", "}\n"]
+          wrapper: ["<%= config.prefix %>", ""]
```
silverwind: What was this `}\n` for?

Mottie: That used to close the main `@-moz-document`.

silverwind: Ah, right so because the closing bracket is now preserved, we drop it here.

Gruntfile.js
```diff
@@ -165,7 +165,7 @@ module.exports = function(grunt) {
     replacement: config.tab
   }, {
     // remove default syntax themes AND closing bracket
-    pattern: /\s+\/\* grunt build - remove to end of file(.*(\n|\r))+\}$/m,
+    pattern: /(\s+\/\* grunt build - remove start \*\/)(.*)(\s+\/\* grunt build - remove end \*\/)$/sm,
```
silverwind: These capture groups are unnecessary, right?

Mottie: It's still needed to remove our inline syntax highlighting themes,

xt0rted: The capture groups in the regex aren't needed, they were added to help with testing it.

github-dark.user.css
```diff
@@ -4639,5 +4639,23 @@
   /*[[syntax-theme]]*/
   /*[[syntax-codemirror]]*/
   /*[[syntax-jupyter]]*/
-
 }
+@-moz-document domain("zuora.com") {
```
silverwind: Maybe add a small comment in `github-dark.css` why this domain is styled. It's not obvious to me.

Gruntfile.js
silverwind: Then I'd say remove them.

github-dark.user.css
xt0rted: The url used in these iframes contains `github.com` which means we can change this to a regex so it's more specific:
```regex
https://www\.zuora\.com.*github\.com.*
```

github-dark.user.css
```diff
@@ -4639,5 +4639,24 @@
...
+  /* Styles the payment form iframe */
```
xt0rted: I had to put this inside the document rule because when the generate task runs it indents it for some reason.

silverwind: Yes, this is because it also runs the [clean task](https://github.com/StylishThemes/GitHub-Dark/blob/4fb5d2a985c2e3aebd440b01447fd0e19ad8a9fc/Gruntfile.js#L424) after generate as the generator would otherwise put all selectors on a single line.

silverwind:
I'd say this is good to merge now.

PR #866, opened 2019-03-02
Update payment iframe match method and colors

xt0rted:
Switches the document rule to use a regex so we don't style the form on other sites or https://www.zuora.com/, and updates the colors to match the rest of the style.

**Before**

image

**After**

image


the-j0k3r:
LGTM =)

silverwind:
Would add some background to the "Pay with" buttons, otherwise looking good.

xt0rted:
Getting those radio buttons to look right with a background color was a real pain. Hopefully my fix doesn't break them in other spots (I couldn't find any).

PR #868, opened 2019-03-04
Add generator rules to fix gist search icons & remove unused selectors

xt0rted:
Wanted some feedback to see if this is how we wanted to handle these, and if there's any rules we might want to remove if we keep it.

image

image

They style the private icon on hover but it looks like 🗑 and it hard to see (even with GHD disabled) so it seemed better to leave it white.

image

This is found under `https://github.com/orgs/:org/people/:user`

image image

There's also a commit that removes some classes that aren't used anymore, and removes the dropdown border color so it uses the autogenerated value which is what's used on the main site's search.


the-j0k3r:
Those private repo locks are looking like retained the original color after your chnages in https://github.com/StylishThemes/GitHub-Dark/pull/825, arent all these lock icons to be same color?

xt0rted:
> arent all these lock icons to be same color?

For some reason they're not consistent with the lock icon. In some places it's gold, in some (as seen here) it's gold with an opacity, and in others it's the default grey.

The generator rule uses our `#cb4` yellow but with the `.5` opacity like they're using. It'd be nice if they could make up their mind on this.

silverwind:
I'd make the yellow a bit more vibrant.

the-j0k3r:
> I'd make the yellow a bit more vibrant.

The yellows we have are far from vibrant, e.g `#cb4` is urine yellow like @mottie would say. I think we should see if better yellow shades can be used instead.

xt0rted:
I just added 995f41879400f1a098599e8d2c1118528d4d2ca0 which is a little bit related to these changes.

When managing installed apps you can pick which repos they have access to, and in the dropdown private repos aren't being styled. They're using a variation of the same color that was added in ae08b2e595c2b886ee3d6a2d51faf5c915f7991e.

The icon has the class `.integrations-install-target .octicon-lock`.

image

I'm not sure if these `rgba()` colors are newer changes, or just colors we've missed but I've been seeing a lot of them this past week while digging through the styles.

the-j0k3r:
We have:
```
.integrations-install-target .navigation-focus .octicon-lock,
.integrations-install-target [aria-selected=true].octicon-lock,
``` and both are set to color: inherit 

nothing else exists slightly related.

**Comment by xt0rted** [comment]
The icons look fine on hover, it's their default state that's broken.

File: github-dark.css [hunk 1]
```diff
@@ -1008,6 +1008,26 @@
   body[class="page-responsive"] .pagination .gap {
     color: #5a5a5a !important;
   }
+  /* auto-generated rule for "color: rgba(27,31,35,.3)" */
+  .btn-outline.disabled, .btn-outline:disabled, .commit .commit-title.blank,
```
the-j0k3r: jj

silverwind:
If you can, remove any existing selectors that are newly covered by these generated rules.

xt0rted:
Will do

github-dark.css
```diff
@@ -1008,6 +1008,26 @@
   body[class="page-responsive"] .pagination .gap {
     color: #5a5a5a !important;
   }
+  /* auto-generated rule for "color: rgba(27,31,35,.3)" */
```
the-j0k3r: test

the-j0k3r:
@xt0rted updates on this?

the-j0k3r:
ping

silverwind:
This needs to be reworked, we now include the first two generators. The third is not covered yet.

Issue #911, opened 2019-04-13
Octotree v3 support <3


xPixv:
* **Browser**: Chromium Edge Dev
* **Operating System**: Windows 10
* **Link to page with the issue**: Any repo
* **Screenshot**:
![image](https://user-images.githubusercontent.com/9652227/56082796-bbefe800-5e4f-11e9-83a7-c6c7c1d6b803.png)

![image](https://user-images.githubusercontent.com/9652227/56082788-98c53880-5e4f-11e9-8831-3cdc59cc26fd.png)

Thank you in advance~

BTW, dark mode is provided in Octotree Pro.

the-j0k3r:
Hello @xPixv! Thank you for reaching out with this request. =)

Im assuming you meant this https://github.com/ovity/octotree ?

From my part, Ive no plans at this time to add support for this, but that's not to say someone else wont, but lets wait for other feedback ;)

Thanks -)

Mottie:
We supported the older version of Octotree, I don't see why we can't support the newer version.

Mottie:
But, yeah... moving it to a separate userstyle would be better.

xt0rted:
The only issue I see with styling this now is there's two versions starting with v3. There's the free open source version and a new paid version with extra features (which includes a dark theme actually). Updating the existing style for this shouldn't be an issue but we'll need to rely on pro users to help with any missing styling there.

xt0rted:
Just pushed up https://github.com/StylishThemes/GitHub-Dark/commit/c6e6d85563c03c273bc3aa4bf372c739a403ac33 which is my first pass at supporting v3. There's some old selectors that probably need removing, but this gets us almost all of the way there.

![image](https://user-images.githubusercontent.com/831974/56084205-b7024880-5dfd-11e9-83f4-aa602b849c4b.png)

xPixv:
Hi @xt0rted ,
I found that Octotree Free and Octotree Pro are TWO extentions.

Free: https://chrome.google.com/webstore/detail/octotree/bkhaagjahfmjljalopjnoealnfndnagc
Pro: https://chrome.google.com/webstore/detail/octotree-pro/fjcahddnekkgihjnjnimgiggdmlgcnbc

Pro have dark mode natively. How about GitHub Dark support Free version only? As long as GitHub Dark do not break Pro dark mode style.

MCOfficer:
That would require [whoever does it] to purchase the pro version, if they even have distinctive differences.
[...]

PR #914, opened 2019-04-17
Add another class to increase specificity of octotree

indocomsoft:
Currently, `.octotree-sidebar.octotree-github-sidebar .octotree-views .octotree-tree-view .jstree-anchor` has the same CSS specificity as the one from octotree itself, as both are defined using `!important` and with the same number of classes. Thus, which style is selected is not deterministic (GitHub-Dark is selected in Chrome, original Octotree is selected in Firefox).

I notice that all `jstree-anchor` are stored inside `jstree`, thus adding `.jstree` selector will increase specificity (since there are higher number of classes than the original Octotree) and be selected.

This is the issue with Firefox noted in #911

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -5874,7 +5874,7 @@
   .octotree-github-sidebar a.octotree-toggle:hover,
   .octotree-github-sidebar a.octotree-opts:hover .settings:before,
   .octotree-github-sidebar a.octotree-opts.selected:hover .settings:before,
-  .octotree-sidebar.octotree-github-sidebar .octotree-views .octotree-tree-view .jstree-anchor,
+  .octotree-sidebar.octotree-github-sidebar .octotree-views .octotree-tree-view .jstree .jstree-anchor,
   .octotree-github-sidebar a.octotree-toggle:not(.octotree-loading):hover,
   .octotree-sidebar.octotree-github-sidebar .octotree-views .octotree-view,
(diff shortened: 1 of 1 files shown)


the-j0k3r:
Hi @indocomsoft and thank you for reaching out with this contribution.

If I understand you correctly, this fixes #911 if so, its worth adding `fixes #911` to the commit line, so when this is merged, this will automatically trigger and close that ticket. See https://help.github.com/en/articles/closing-issues-using-keywords for more information.

Just edit your local branch and force push the chnages to update this PR automatically, no need to re-open another ;-)

I'll pass this to @xt0rted to test and decide though. =)

Thanks

indocomsoft:
Thank you for the quick response! I have changed the commit message. Let me know if it can be improved further!

xt0rted:
There's no need to edit the commit message, you can add `fixes ####` to the PR comment and it'll close the issue as well.

the-j0k3r:
> There's no need to edit the commit message, you can add `fixes ####` to the PR comment and it'll close the issue as well.

Its my preference and reason is because it is in commit message title. I look at commit history often and refer to see and navigate from there, Doing in PR comment only closes ticket commit tittle wont be as effective then. (at least it was the behavior unless its changed)

xt0rted:
Thanks for the contribution @indocomsoft

PR #920, opened 2019-04-19
resources site: add initial manual support

the-j0k3r:
see internal discussion

Its important that this doesnt break what already works, I tried adding prefixes but that doesnt apply consistently for this site pages.

Dont merge until we can properly test hence wip label =)

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -7078,3 +7078,347 @@
     fill: #fff;
   }
 }
+@-moz-document domain("resources.github.com") {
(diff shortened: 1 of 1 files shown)


the-j0k3r:
force pushed based on current master at 22339d71899ed603255329be56fd94f3e5d42337

@silverwind

the-j0k3r:
updated based on master at 40e0d3d183e23993ea7ecdef983ab42e16c59dff

PR #921, opened 2019-04-21
Brighten octicons in markdown bodies

Aced14:
**Example:**
https://help.github.com/en/articles/commenting-on-a-pull-request

**Screenshots...**

**Before:**
octicon-before

**After:**
octicon-after


github-dark.css
```diff
@@ -4561,7 +4561,7 @@
   .site-header-nav .dropdown-menu a.active,
   .site-header-nav .dropdown-menu a .octicon-check,
   .marketplace-product-callout .branch-action-item-icon,
-  .user-status-org-button span.text-bold {
+  .user-status-org-button span.text-bold, .markdown-body .octicon {
```
the-j0k3r: This is not needed =)

github-dark.css
```diff
@@ -4810,7 +4810,7 @@
   .delete-topic-button {
     color: inherit !important;
   }
-  svg:not(.octicon):not([fill="none"]) {
+  svg:not(.octicon):not([fill="none"]), .markdown-body .octicon {
     fill: currentColor !important;
```
the-j0k3r: That takes care of all the octicons on that page just fine =)

Mottie:
What's actually needed is this change:

```css
span.octicon svg {
    fill: currentColor;
}
```

Since the `octicon` class is added to a span in this case.

github-dark.css
Aced14: I'll modify this line with the selector @Mottie suggested.

github-dark.css
Aced14: Without this, won't octicon colours change from ``#dcdcdc`` to ``#bebebe``?

In help pages, octicons seem to always be used alongside bolded text (which use ``#dcdcdc``) or used on their own in place of where bolded text would've usually been. A similar colour difference also exists in GitHub's default theme (where octicons are ``#000000`` and normal/bolded text is ``#24292e``).

Removing this line's change would deviate from that colour pattern.

Thoughts?

Aced14:
Just updated this PR.

github-dark.css
the-j0k3r: Add like below then, `markdown-body span.octicon svg`

the-j0k3r:
@Aced14 I had a similar change in https://github.com/StylishThemes/GitHub-Dark/commit/0a5f5e33667cc687cbc2344e455adc59a85835fe for one octicon, so updated that to match all instead.

silverwind:
Can you limit this style to the help domain? Main page has `fill: currentColor`, it's only absent on the help page so `fill` defaults to browser default `#000`.

PR #927, opened 2019-04-24
Mobile commit header fixes

xt0rted:
Wanted to PR this to see if someone sees any issues with the change or if it may break something I'm unaware of.

The generated rules are styling `background-color` and `border-color` so this change follows that. It also removes `!important` which was another recent change.

Switching to style just the color removes the left & right border that's showing on the blue section, and the removal of the transparent selector is because it's hiding the border between the breadcrumb section and the blue section. On desktop the breadcrumb section doesn't show on this page, but it does when viewing a [single file](https://github.com/StylishThemes/GitHub-Dark/blob/master/package.json) (the `GitHub-Dark / package.json` text) but there's no border there on desktop.

This is the page used for the screenshots https://github.com/StylishThemes/GitHub-Dark/commit/7cf20aa6ce2815af23bd466be49bcdca2edb773d.

**Before**

github com_StylishThemes_GitHub-Dark_commit_7cf20aa6ce2815af23bd466be49bcdca2edb773d(Pixel 2)

**After**

github com_StylishThemes_GitHub-Dark_commit_7cf20aa6ce2815af23bd466be49bcdca2edb773d(Pixel 2) (1)

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -3266,7 +3266,7 @@
   }
   .gisttabs a:not(.selected), .gisttabs a:not(.selected):hover,
   .timeline-comment-wrapper, .avatar-stack .avatar, .full-commit .commit-meta,
-  .breadcrumb, .tabs > a, .header .divider-vertical, .about-menu-link {
+  .tabs > a, .header .divider-vertical, .about-menu-link {
     border-color: transparent !important;
   }
   /* darken border */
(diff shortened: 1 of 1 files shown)


xt0rted:
Just found the first casualty of the `border-color` change.

image

xt0rted:
Turns out the blue section wasn't even using the blue border so I reverted that change and removed the selector from that rule. That fixes the branch labels and the original issue.

xt0rted:
It's too early. The fix was actually just adding `:not(.mobile)` to the original selector. I'll leave this for now so someone else can verify I'm not breaking anything.

PR #931, opened 2019-04-29
Fixed repo private lock icon color.

AM-77:
Added !important to force the style rule to be applied to the classes assigned to it.


silverwind:
Sorry, this is not the right solution (`!important` almost never is). Also, this modifies a autogenerated section of the style, so the change would be lost the next time we regenerate it.

PR #943, opened 2019-05-10
Setup npm scripts for common tasks

xt0rted:
It's my understanding that you don't need to use `npx` when calling Grunt from an npm script. I don't have Grunt installed globally, it's only installed as a dependency in this repo, and using these scripts works perfectly fine for me.

I'm PRing to see what the build server says and for some feedback from @StylishThemes/github to see if anyone has any issues running these.

Task | Command
-- | --
clean | `npm run clean`
generate | `npm run generate`
test | `npm run test` or `npm test`
usercss | `npm run usercss`

We could add more of these in for doing releases, I just wasn't sure what one(s) we'd want right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,10 @@
     "node": ">=10"
   },
   "scripts": {
-    "test": "npx grunt lint"
+    "clean": "grunt clean",
+    "generate": "grunt generate",
+    "test": "grunt lint",
+    "usercss": "grunt usercss"
   },
   "devDependencies": {
     "css": "2.2.4",


silverwind:
Yes, build script have always worked without `npx`. I just thought of adding it for the ability to copy the string as-is to the shell.

Mottie:
If you're going to do this, then shouldn't we also update the readme?

https://github.com/StylishThemes/GitHub-Dark#development

the-j0k3r:
Can we have aliases that dont need the word run?

xt0rted:
The only built-in alias is `test`. `build` has a pre-existing use so that needs to be called as `npm run build` which is why I didn't bother adding it in (yet).

silverwind:
> update the readme

calling grunt scripts should still be done with `npx`. It just happens that it can be ommited in package scripts as those automatically use module binaries. Still, I'd say we could also just leave it as is, don't see much benefit of removing npx in the scripts, except it might confuse uninformed users.

PR #1010, opened 2019-10-08
Use a fail fast build matrix

xt0rted:
Setting fail-fast to false will let each OS run to completion even if one of them fails. This will let us see if an issue is OS specific or not.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,44 +3,13 @@ name: CI
 on: push
 
 jobs:
-  build-macOS:
-    runs-on: macOS-latest
+  build:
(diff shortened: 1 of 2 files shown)


xt0rted:
@silverwind @Mottie @the-j0k3r this switches back to a build matrix now that I see there's a fail fast option we can disable (it's the reason I didn't go this route originally). It also removes the travis config assuming we're ok getting rid of it now. Just looking for the 👍  or 👎 on this before I merge.

the-j0k3r:
Real life must have taken some ppl hostage. I say lets give it a whirl.

PR #1018, opened 2019-10-18
Add graphql explorer to the generator

xt0rted:
I was playing around with this page earlier and found a number of unstyled parts due to a style sheet that wasn't being included in the generator.

I added the url in and started adding some of the missing colors but wanted to PR the changes to see if this was adding too many new selectors/rules in vs. manually styling it. There's some manual styles that can be removed but I didn't get to those yet.

This is still a work in progress but I haven't seen any issues elsewhere on the site from it.

### DIFF
--- a/tools/generate.js
+++ b/tools/generate.js
@@ -293,7 +293,11 @@ const sources = [
   {url: "https://github.com"},
   {url: "https://gist.github.com"},
   {url: "https://help.github.com"},
-  {url: "https://graphql-explorer.githubapp.com/"},
+  {
+    url: "https://graphql-explorer.githubapp.com/", 
+    prefix: ".graphql-container",
+    match: [".graphql-container"],
+  },
   {url: "https://developer.github.com", prefix: "html[prefix]"},
   {
     url: "https://www.githubstatus.com/",
--- a/tools/generate.js
+++ b/tools/generate.js
@@ -17,7 +17,9 @@ let mappings = {
   // ==========================================================================
   // Background
   // ==========================================================================
+  "$background: white": "#181818", // graphql explorer
   "$background: #fff": "#181818",
+  "$background: #ffffff": "#181818", // graphql explorer
   "$background: #ffe": "#242424",
   "$background: #fdfdfd": "#1c1c1c",
   "$background: #fafbfc": "#181818",
@@ -57,6 +59,8 @@ let mappings = {
   "$border: #959da5": "#484848",
   "$border: #c3c8cf": "#484848",
   "$border: #dfe2e5": "#343434",
+  "$border: #d6d6d6": "#343434", // graphql explorer
+  "$border: #d3d6db": "#343434", // graphql explorer
   "$border: #d1d5da": "#404040",
   "$border: #ddd": "#343434",
   "$border: #e1e4e8": "#343434",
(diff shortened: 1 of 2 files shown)


tools/generate.js
```diff
@@ -158,6 +164,7 @@ let mappings = {
 
   "color: #0366d6": "color: /*[[base-color]]*/ #4f8cc9", // needs to be after #333
   "color: #1074e7": "color: /*[[base-color]]*/ #4f8cc9",
+  "color: #1F61A0": "color: /*[[base-color]]*/ #4f8cc9", //  graphql explorer
```
xt0rted: @silverwind are these values case sensitive? A few of the colors in this style sheet use all caps which is why they're using it here too.

the-j0k3r: I dont change the hex case from the default stylesheets when adding them to generator, though I always use lower case for replacement colors as its easier to read 8 vs B etc.

xt0rted:
Looks like I found the first big issue with these changes. The style sheet used on this domain has classes with colors that don't match github.com now. Very specifically the `.link-gray` class uses `#767676` while on github.com it's `#586069`. This causes two auto-generated rules to be added and the one from the graphql explorer takes priority which darkens the text in a number of spots.

image

image

the-j0k3r:
> The style sheet used on this domain has classes with colors that don't match github.com now. Very specifically the .link-gray class uses #767676 while on github.com it's #586069. This causes two auto-generated rules to be added and the one from the graphql explorer takes priority which darkens the text in a number of spots.

What do you think about these sorts of additions going to their own @moz-rule so as to avoid these conflicts? I did the same with another section a while back for same reasons though I used the autogenerated rules to build the specific section.

Thats until the generator can handle those sections automatically.

xt0rted:
It looks like that's what'll have to happen here too. Once everything is in place I can copy all the rules into their own section, then remove the colors from the generator for now.

tools/generate.js
silverwind: They are compared insensitively, thought I do prefer lowercase in this file.

tools/generate.js
```diff
@@ -283,6 +293,7 @@ const sources = [
   {url: "https://github.com"},
   {url: "https://gist.github.com"},
   {url: "https://help.github.com"},
+  {url: "https://graphql-explorer.githubapp.com/"},
```
silverwind:
```suggestion
  {
    url: "https://graphql-explorer.githubapp.com/", 
    prefix: ".graphql-container",
    match: [".graphql-container"]
  },
```

silverwind:
Should be scoped to a selector like above. html and body don't have any usable properties, so `.graphql-container` seems the best we can do.

tools/generate.js
the-j0k3r: @xt0rted commit @silverwind suggestion and regenerate, if it works (it should just like mobile styles go ahead and merge.

the-j0k3r:
@xt0rted rebase on master and regenerate I committed @silverwinds suggestion in https://github.com/StylishThemes/GitHub-Dark/pull/1018/commits/023830c632d291c3c008d5f3741c6390842d75a8

tools/generate.js
```diff
@@ -283,6 +293,11 @@ const sources = [
   {url: "https://github.com"},
   {url: "https://gist.github.com"},
   {url: "https://help.github.com"},
+  {
+    url: "https://graphql-explorer.githubapp.com/",
```
the-j0k3r:
```suggestion
    url: "https://graphql-explorer.githubapp.com/",
```

the-j0k3r:
Ive moved this out of draft so to get it moving.

silverwind:
Looking at it again, we might be able to use `html[prefix*="ogp.me"]` for the prefix. That URL refers to some graph protocol, so should be unique enough I guess.

silverwind:
Also, link to it because we don't have one yet in this thread: https://developer.github.com/v4/explorer

silverwind:
Implemented this. Not really tested as the actual application looks broken to me (just shows `Loading...`).

PR #1049, opened 2020-01-08
Add support for font-feature-settings on the code font

xt0rted:
Fonts like [Fira Code](https://github.com/tonsky/FiraCode#stylistic-sets) and [Inter](https://rsms.me/inter/#features) (the font used by GitHub on their marketing pages) support feature settings. I use Fira Code in my editors and more recently started using on GitHub for consistency. The missing part was being able to turn some of these features on like I can in VSCode. I've been running this change for some time now in its own style with no issues, but I wanted to PR it in case there was something I was missing when adding a new setting to the theme.

The default value is set to `normal` which is the browser default, but if you change it to something like `'ss02', 'ss03', 'ss04'` when using Fira Code you'll notice a difference in how `<=`, `&`, and `$` are displayed.

More info on the setting:

- https://developer.mozilla.org/en-US/docs/Web/CSS/font-feature-settings
- https://css-tricks.com/almanac/properties/f/font-feature-settings/

### DIFF
--- a/github-dark.user.css
+++ b/github-dark.user.css
@@ -194,6 +194,7 @@
   } EOT;
 }
 @advanced text font-choice "Code Font" "Menlo"
+@advanced text font-features "Code Font Features" normal
 @advanced text font-size-choice "Code Font Size" "default (eg 10px)"
 @advanced dropdown code-wrap "Code Wrap" {
   nowrap "No Wrap" <<<EOT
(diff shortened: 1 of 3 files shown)


github-dark.user.css
```diff
@@ -194,6 +194,7 @@
   } EOT;
 }
 @advanced text font-choice "Code Font" "Menlo"
+@advanced text font-features "Code Font Features" normal
```
silverwind: Should `normal` be quoted?

xt0rted: In actual css you don't quote `normal`, but in this instance it looks to work either way and the browser gets `normal` instead of `"normal"`. `base-color` doesn't use quotes, but all the others do. Want me to change them so they're consistent?

silverwind: Yeah, I'd like those values to reflect what goes into the CSS so make them unquoted where they appear unquoted in CSS. I think only `bg-custom` currently violates this. Would appreciate if you could test those variables after the change.

silverwind:
Merged it, we can follow up with fixing the other options later.

Issue #1056, opened 2020-02-07
PR Toolbar shadow does not stick to top and obstructs view


KnutZuidema:
* **Browser**: Firefox 72
* **Operating System**: 5.4.15-arch1-1
* **Link to page with the issue**: https://github.com/StylishThemes/GitHub-Dark/pull/959/files
* **Screenshot**:
![image](https://user-images.githubusercontent.com/28651746/74030394-3f666100-49af-11ea-9d4b-b9e1d31f82af.png)

the toolbar shadow somehow does not stick to the top of the page

the-j0k3r:
@KnutZuidema Thanks for this report.

Please ensure you have no other styles for GitHub Interfering here. As is I cant see this issue,
Also make sure you have the latest GitHub Dark v1.22.203 installed,

![Screenshot_2020-02-07 Codemirror tabs by Mottie · Pull Request #959 · StylishThemes GitHub-Dark](https://user-images.githubusercontent.com/31389848/74031105-f4e5e400-49b0-11ea-845b-04d741155c1e.png)

If you have other information how to duplicate this, please let me know.

I'm assuming this is not a OS issue, but in any case In Windows 10 is as you see above screenshot.

the-j0k3r:
OK was able to duplicate. Required to expand file in order to trigger it.

![generic](https://user-images.githubusercontent.com/31389848/74031729-8d309880-49b2-11ea-91f9-c4c797bdbaaa.gif)

KnutZuidema:
BTW, in case this was not clear, this happens for every PR, not just for the link I posted. I believe it is because of some inline css which adds a `top: 60px !important;` (at least it fixes it if I remove that) but as far as I can tell this is added server side by github.

the-j0k3r:
This is fixed by
```css
  .toolbar-shadow[style="top: 60px !important;"] {
    margin-top: -60px !important;
  }
```

Yes theres an inline style but am not sure if the above is best fix for this.

@silverwind @xt0rted for feedback

KnutZuidema:
I can confirm that this fixes it for me as well

silverwind:
Would use `.toolbar-shadow[style*="top: 60px"]`

PR #1064, opened 2020-02-13
fix toolbar shadow offset

### LINKED ISSUE #1063: Toolbar shadow overlays PR diff
* **Browser**: Firefox
* **Operating System**: Arch Linux
* **Link to page with the issue**: -
* **Screenshot**: details hidden, the purple does not show up in the browser

image

KnutZuidema:
In the PR diff view when using the new notifications beta, the toolbar shadow is offset by 133px. With this style it causes the shadow to overlay the diff view.
This fix is equivalent to the fix for issue #1056
fixes #1063

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -9121,6 +9121,9 @@
   .toolbar-shadow[style*="top: 60px"] {
     margin-top: -60px !important;
   }
+  .toolbar-shadow[style*="top: 133px"] {
+    margin-top: -133px !important;
+  }
   .gh-header-shadow {
     border: none !important;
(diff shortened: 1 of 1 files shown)


the-j0k3r:
thanks

xt0rted:
I just looked into this and if we remove all of the existing tweaks the sticky toolbars display correctly. I just pushed these changes up in 75229bb and did a new release. Give v1.22.214 a try and let me know if you find any other issues with these.

KnutZuidema:
Looks good to me. :+1: That's much cleaner and probably more future proof

Issue #180, opened 2014-09-18
Add rule to edit CSS for awesome extension Octotree


zckrs:
Hi, 
If you don't know check here : [Octotree](https://chrome.google.com/webstore/detail/octotree/bkhaagjahfmjljalopjnoealnfndnagc)

Is possible to add a style for a dark Octotree ?

Issue #170, opened 2014-09-03
Background for split diffs is bright white


michaelmior:
On the new split diffs, the background colour used for whitespace is a bright white. This should probably just match the normal background text colour.

 ![splitdiff](https://cloud.githubusercontent.com/assets/82501/4141121/787dafe0-33ab-11e4-8256-1a67937b2f4f.png)

PR #1069, opened 2020-02-21
Green button revamp

the-j0k3r:
@silverwind here is a slightly tweaked version which uses more of already existing colors rather than adding new ones.

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -5420,13 +5420,13 @@
   body[class="page-responsive"] .btn-primary,
   body.zh_logged_in .zh-webapp.zh-epic-creator-page-visible .btn-primary,
   html.ghh-theme-classic .ghh .ghh-state-verified {
-    background-color: #061;
+    background-color: #373;
   }
   /* remap-css rule for "background-color: #28a745 !important" */
   .bg-green, .bg-green-5, #graphiql .bg-green, #graphiql .bg-green-5,
   body.status .bg-green, body[class="page-responsive"] .bg-green,
   body[class="page-responsive"] .bg-green-5 {
-    background-color: #061 !important;
+    background-color: #373 !important;
   }
   /* remap-css rule for "background-color: #2cbe4e" */
   .State--green, .diffstat-block-added, .progress-bar .progress,
@@ -5437,7 +5437,7 @@
   body[class="page-responsive"] .progress-bar .progress,
   html.ghh-theme-classic .ghh .ghh-state-open,
   .ghh-theme-github .ghh-reviews .ghh-state-icon-success {
-    background-color: #061;
+    background-color: #373;
   }
   /* remap-css rule for "background: #dcffe4" */
   .boxed-group-list > li.selected {
--- a/github-dark.css
+++ b/github-dark.css
@@ -9163,22 +9163,24 @@
   /* interactive button - bright border */
   .btn-primary:not(.compare-button):not(.btn-outline):not([disabled]):not(.disabled),
   .button.primary, .btn-theme-green, .btn-primary-mktg {
-    background: linear-gradient(#407045, #305530) !important;
-    border-color: #083 !important;
+    background: linear-gradient(180deg, #373 5%, #060 100%) !important;
+    border-color: #193 !important;
   }
-  .button.primary:disabled, .button.primary:disabled:hover,
-  .button.primary.disabled, .button.primary.disabled:hover,
-  .btn-primary:disabled, .btn-primary:disabled:hover, .btn-primary.disabled,
-  .btn-primary.disabled:hover, .btn-theme-green:disabled,
-  .btn-theme-green:disabled:hover {
-    background: linear-gradient(#203522, #152715) !important;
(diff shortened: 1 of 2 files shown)


silverwind:
Please limit changes to just `github-dark.css`

the-j0k3r:
Was just for testing - removed

silverwind:
Hmm, I prefer your previous (more vibrant) colors more. Maybe use those and possibly adjust existing colors to match if they are reasonably close?

the-j0k3r:
done

the-j0k3r:
Also we should change   /* auto-generated rule for "background-color: #2cbe4e" */ for `#337733` `#373` and it will match the new buttons

silverwind:
Yeah let's update the generator here as well.

the-j0k3r:
Actually a few more `#061` need to be `#373`

 /* auto-generated rule for "background-color: #28a745 !important" */

generic

IDK if there are other places

the-j0k3r:
thats should take care of the 3

https://github.com/StylishThemes/GitHub-Dark/pull/1069/commits/125f9a73a6cd7e97f77b3120857dd51ddedd2f0a

silverwind:
Used this for a bit. May be just me but I'd go one tiny shade darker on the color.

the-j0k3r:
That was what I had initially.

the-j0k3r:
what about instead of `#337733` we go for `#117700`

silverwind:
Hmm yes, `#170` could work too I think.

the-j0k3r:
OK pushed changes =) I really liky :drooling_face:

github-dark.css
```diff
@@ -9105,30 +9105,35 @@
...
   }
   /* Green hovered */
-  .noproblems .pageheader_wrap, a.new-repo:hover {
-    background: #373 !important;
-  }
   .btn-primary:not(:disabled):not(.disabled):not(.compare-button):not(.btn-outline):hover,
   .button.primary:not(:disabled):hover, .btn-theme-green:hover,
   .btn-primary-mktg:hover {
-    background: linear-gradient(#508055, #407045) !important;
+    background: linear-gradient(180deg, #170 5%, #170 100%) !important;
```
silverwind:
```suggestion
    background: linear-gradient(180deg, #280 5%, #170 100%) !important;
```

Hover should not remove gradient.

silverwind:
Hmm I'm not 100% sure yet. It does look okay on one of my screens, but too flashy on another 😱

the-j0k3r:
Yea I have same issue with my laptop screen, the colors are not same as in my desktop screen

Desktop screen represents the colors faithfully Its calibrated IPS screen, while laptop screen is not anywhere close, I its a TN screen.

Overall these offer the best balance.

github-dark.css
the-j0k3r: Thats far too much, even `#180` is already too much

the-j0k3r: How about
```suggestion
    background: linear-gradient(180deg, #180 1%, #170 70%) !important;
```

the-j0k3r: lol whats with the orange

the-j0k3r: <s>lol whats with the orange<\s>

Never mind GH is misbehaving.

the-j0k3r:
Ive just also noticed we dont have any

```
.btn-primary.selected, .btn-primary:active, .btn-primary[aria-selected="true"], [open] > .btn-primary {
	background-color: #279f43;
	background-image: none;
	border-color: rgba(27,31,35,.5);
	box-shadow: inset 0 .15em .3em rgba(27,31,35,.15);
}
```

generic

silverwind:
I'm sorry but can you restore your previous iteration once more? (With the gradient fix if possible).

the-j0k3r:
what `#373`?

the-j0k3r:
done, though this isnt better than before, hover gradient is either too bright or too dark on top or bottom or both, also, I think its best to use the temp usercss to test colors, this back and forth is consuming too much time.

silverwind:
> this back and forth is consuming too much time.

you better get used to the branching model 😆

the-j0k3r:
> you better get used to the branching model laughing

It was what was done here =) I meant making changes and not adding silly commits for each change

xt0rted:
The primary buttons seem too bright now. They stand out from the other greens more than the old colors did and they're the brightest thing on the page. The old colors seemed better here.

silverwind:
Don't think you'll get used to them? I think the old ones looked a bit "dull".

xt0rted:
They just feel too vibrant compared to the rest of the color palette.

silverwind:
Maybe we can make them a bit duller, but not as dull as before 😉

the-j0k3r:
> Maybe we can make them a bit duller, but not as dull as before wink

ugh lol

silverwind:
I don't think many users would want to configure button colors 😆

the-j0k3r:
I dont do styles for others, though they a re welcome to decide to use them or not.

silverwind:
Will likely do a slight revision of the color to desaturate it a bit.

PR #1078, opened 2020-03-01
Remove unneeded quotes from css values

xt0rted:
This is based on the conversation in https://github.com/StylishThemes/GitHub-Dark/pull/1049#discussion_r364920445. I'd like someone else to test this out before it gets merged. In Chrome I didn't see any issues with this change.

### DIFF
--- a/tools/usercss-template.css
+++ b/tools/usercss-template.css
@@ -28,7 +28,7 @@
   none "None" <<<EOT
   none EOT;
 }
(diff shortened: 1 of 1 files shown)


the-j0k3r:
Well this removes a few quotes from the usecss option values, I cant see any ill effects about this.

Also this one https://github.com/StylishThemes/StackOverflow-Dark/blob/master/tools/usercss-template.css has none also.

Trivial change I would say.

PR #1081, opened 2020-03-03
Remove manual style of selected search result text

### LINKED ISSUE #1080: Selected search result text looks darker then it should be
I'm not sure if it's just me or if something recently changed but the text in the search box looks darker then it should be now. Right now it's `#dcdcdc` because of a manual style but other areas that use the primary color are all set to use `#ffffff` for the text.

Should the manual style be removed so the value from the generator is used?

Using `#dcdcdc`

image

Using `#ffffff`

image

xt0rted:
Fixes #1080


the-j0k3r:
Well Im not sure about that, see the comment on issue. Ill test with this though.

the-j0k3r:
See the [comment on issue](https://github.com/StylishThemes/GitHub-Dark/issues/1080#issuecomment-593791339). Ill test with this though that comment is without this PR

github-dark.css
```diff
@@ -8716,10 +8716,8 @@
   .counter, .Counter, .state, .State, .dashboard-notice h2,
   .site-header-dark .site-header .site-header-link, .version-picker a:hover,
   .markdown-format .octicon-link, .Box-row--hover-blue.link-gray:hover,
-  .jump-to-suggestions-results-container .navigation-focus .jump-to-suggestions-path,
   .jump-to-suggestions-results-container .navigation-focus .jump-to-octicon,
   .jump-to-suggestions-results-container mark,
```
the-j0k3r:
```suggestion
```

needs needs to go away

github-dark.css
```diff
@@ -8726,7 +8725,8 @@
   .site-header-nav .dropdown-menu a.active,
   .site-header-nav .dropdown-menu a .octicon-check,
   .marketplace-product-callout .branch-action-item-icon,
-  .user-status-org-button span.text-bold, .octicon-fold {
+  .user-status-org-button span.text-bold, .octicon-fold,
+  .jump-to-suggestions-path {
```
the-j0k3r: Either this or something else, but this looks more balanced to me between highlight and normal.

silverwind:
Some additional removals done in https://github.com/StylishThemes/GitHub-Dark/commit/836f868410e83acc6c6e111887a63053a29fefd3. They were all overwriting styles already in the generated rules.

I sometimes feel tempted to just nuke all of these manual rules and see how much really breaks.

xt0rted:
I've been wanting to do that for awhile now. We should try to reorganize it so things like marketing pages, extensions, beta features, and GHE are grouped in their own sections so we can easily purge them without worrying about breaking the main parts of the site. It's nice to reuse rules but it makes cleaning things up a lot harder.

If the generator was made so it didn't merge rules and left them as they were from the source files, in the same order, and withe media queries, wouldn't that help alleviate a lot of the custom rules we have to add in? Basically copy the css file, find & replace with the generator, and then delete everything that doesn't set a color.

silverwind:
Yes, separate per-site section should be doable pretty easily by running the tooling multiple times.

Media queries would be harder, I opened https://github.com/silverwind/remap-css/issues/1 for this.

silverwind:
> same order

And that's https://github.com/silverwind/remap-css/issues/2. The issue with original order is that it will blow up the line count dramatically, like 4-5 times the current amount.

xt0rted:
We'll just have to split each site/subdomain out into its own file and then bundle everything as a browser extension 😜

silverwind:
Thing is I've been rejecting the idea of a build process mostly just because I'm so used to copying the CSS to Stylus, so not sure what to do. Stylus does not support `@import`, so I guess we'll have to live on with that huge file.

the-j0k3r:
> I sometimes feel tempted to just nuke all of these manual rules and see how much really breaks.

Feel free to nuke everything and then fix it! lol

PR #1091, opened 2020-03-30
Fix tinted background color option

fregante:
This PR is untested, but the screenshot is taken with a real reproduction via the custom background

# Before

before
# After
after

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -6839,7 +6839,7 @@
   }
   /* end remap-css rules */
   body {
-    background-color: #222 !important;
+    background-color: #202020 !important;
     background-image: var(--bg-img) !important;
     background-clip: border-box !important;
     background-origin: padding-box !important;
--- a/github-dark.css
+++ b/github-dark.css
@@ -11622,7 +11622,7 @@
   }
   body {
     color: #e0e0e0;
-    background-color: #222;
+    background-color: #202020;
     background-image: var(--bg-img);
     background-clip: border-box;
     background-origin: padding-box;
--- a/tools/usercss-template.css
+++ b/tools/usercss-template.css
@@ -26,7 +26,7 @@
   custom "Custom" <<<EOT
   var(--bg-custom) EOT;
   none "None" <<<EOT
-  #202020 EOT;
+  none EOT;
 }
 @advanced text bg-custom "Custom Background" {{image}}
 @advanced dropdown bg-options "Background image type" {
--- a/tools/usercss-template.css
+++ b/tools/usercss-template.css
@@ -26,7 +26,7 @@
   custom "Custom" <<<EOT
   var(--bg-custom) EOT;
   none "None" <<<EOT
-  none EOT;
+  #202020 EOT;
 }
 @advanced text bg-custom "Custom Background" {{image}}
 @advanced dropdown bg-options "Background image type" {


the-j0k3r:
Hi @fregante adding color here actually breaks the intended design of this. If you must change something its at the body entry.

https://github.com/StylishThemes/GitHub-Dark/blob/a4c96edabb6c015aeb6c29d52f85a17c04dd496c/github-dark.css#L11625

So instead of `#222222` change it to your `#202020` Which is what none will mean.

tools/usercss-template.css
```diff
@@ -26,7 +26,7 @@
   custom "Custom" <<<EOT
   var(--bg-custom) EOT;
   none "None" <<<EOT
-  none EOT;
+  #202020 EOT;
```
the-j0k3r: Hi @fregante adding color here actually breaks the intended design of this. If you must change something its at the body entry.

https://github.com/StylishThemes/GitHub-Dark/blob/a4c96edabb6c015aeb6c29d52f85a17c04dd496c/github-dark.css#L11625

So instead of `#222222` change it to your `#202020` Which is what none will mean.

xt0rted:
Is this meant to adjust the background so the tabs match the rest of the page when not using a background image? If so I've been running this in a separate style for this very reason.

```css
/* Tabs - set color to BG color so they blend fully */
.pagehead-tabs-item.selected,
.reponav-item.selected,
.tabnav-tab.selected {
  background-color: #222 !important;
  border-bottom-color: #222 !important;
}
/* pending review comment tabs */
.is-pending .tabnav-tab.selected {
  background-color: #181818 !important;
}
/* https://github.com/settings/replies */
.saved-reply-form .tabnav-tab.selected {
  background-color: #181818 !important;
  border-bottom-color: #181818 !important
}
```

fregante:
I don’t think all of that is needed, in my case I just had to change the main background color once

xt0rted:
Yea, it looks like some of this has since been updated. The pending review comment is still needed but the saved reply rule isn't needed at all now

fregante:
@the-j0k3r like this?

silverwind:
Changing body is fine with me. I want these tabs to somewhat work with the default textured background (thought CSS can not cleanly fix that).

image

xt0rted:
This change is inside of `@-moz-document domain("stylishthemes.github.io") {` which doesn't seem right to me.

github-dark.css
```diff
@@ -11622,7 +11622,7 @@
   }
   body {
     color: #e0e0e0;
-    background-color: #222;
+    background-color: #202020;
```
the-j0k3r: Yes :+1:

the-j0k3r:
> This change is inside of `@-moz-document domain("stylishthemes.github.io") {` which doesn't seem right to me.

that also needed to be changed.

See https://github.com/StylishThemes/GitHub-Dark/pull/1091#discussion_r400697133

the-j0k3r:
> Is this meant to adjust the background so the tabs match the rest of the page when not using a background image? If so I've been running this in a separate style for this very reason.
> 
> I'm not sure if this is up to date still, been awhile since I last checked.
> 
> ```css
> /* Tabs - set color to BG color so they blend fully */
> .pagehead-tabs-item.selected,
> .reponav-item.selected,
> .tabnav-tab.selected {
>   background-color: #222 !important;
>   border-bottom-color: #222 !important;
> }
> /* pending review comment tabs */
> .is-pending .tabnav-tab.selected {
>   background-color: #181818 !important;
> }
> /* https://github.com/settings/replies */
> .saved-reply-form .tabnav-tab.selected {
>   background-color: #181818 !important;
>   border-bottom-color: #181818 !important
> }
> ```

Perhaps only the #222 ones as clearly the rest are separate colors on purpose.

xt0rted:
The `.is-pending .tabnav-tab.selected` rule fixes the tabs on the pending pr review ui.

image

The `.saved-reply-form .tabnav-tab.selected` rule doesn't look to be needed anymore, the ui there changed slightly and is styled correctly now.

fregante:
> The `.is-pending .tabnav-tab.selected` rule fixes the tabs on the pending pr review ui.

That's an unrelated issue, it appears even with the regular background. Please open another issue

fregante:
@silverwind is there anything else that needs to be fixed?

the-j0k3r:
> The `.is-pending .tabnav-tab.selected` rule fixes the tabs on the pending pr review ui.

Yep unrelated area.

PR #1093, opened 2020-04-11
Fix code snippet background in security alerts

xt0rted:
Since this uses `:not()` I wanted to PR it so @silverwind can approve the use of it or tell me how horrible this is 😃

This can be seen under the [Security > Alerts](https://github.com/StylishThemes/GitHub-Dark/network/alerts) section of a repo with an open alert. I'm not sure if it's an issue for all syntax themes, I'm using Monokai here.

Before | After
-- | --
github com_xt0rted_dotnet-format_network_alert_package-lock json_minimist_open (1) | github com_xt0rted_dotnet-format_network_alert_package-lock json_minimist_open

### DIFF
--- a/github-dark.css
+++ b/github-dark.css
@@ -6925,7 +6925,7 @@
     background-image: none !important;
   }
   /* end remap-css rules */
-  body {
+  body, body.bg-gray {
     background-color: #202020 !important;
     background-image: var(--bg-img) !important;
     background-clip: border-box !important;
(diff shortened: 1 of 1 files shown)


silverwind:
Those two existing `:not`s need to go. Now if only I knew where `body.bg-gray` is an issue 😉

silverwind:
Suggestion: Remove the `not`s and add `body.bg-gray, .team-discussions-container` to `background: transparent` instead, leaving only the selector `.bg-gray` behind.

xt0rted:
The `body.bg-gray` is used on the app installation page (7215b7b2c9aff47f22ca8786c87bf72e16cba2ce) and the `.bg-gray.team-discussions-container` is used on https://github.com/orgs/StylishThemes/teams/github (775d4f13dd67484a1e73d52877d83d47350ff972).

```html
<body class="logged-in env-production min-width-lg bg-gray">...</body>
```

```html
<div class="team-discussions-container d-flex flex-wrap bg-gray">...</div>
```

github com_apps_imgbot_installations_new_permissions_target_id=831974(iPad)

silverwind:
Good digging. Does it make those backgrounds  transparent with my suggestions? If yes, go ahead.

xt0rted:
The `.team-discussions-container` is already being explicitly styled and wasn't needed here, and the `body.bg-gray` needs to have the background added to it otherwise it shows up white. That left a plain `.bg-gray` selector which is already being handled by the generator so that was able to go too.

PR #1181, opened 2020-07-16
docs: Fix ZenHub link

goodjack:


the-j0k3r:
This also fixes a security warning when entering https://www.zenhub.io/

thanks not that I ever visit that page until now.

Issue #1154, opened 2020-06-28
button primary green/gray color + disabled + selected states


the-j0k3r:
* **Browser**: FF 77.0.1 (64-bit)
* **Operating System**:  Windows 10 pro
* **Link to page with the issue**:  ANy with green /gray buttons
* **Screenshot**:

# Issue 1 green buttons color

The green buttons are now too bright way off the balance between green/red yellow
The OPEN state color is aberrantly off the right bg color.

```css
State--green {
    background-color: #595;
```

this use to be set at same color as green buttons its now WAY OFF as well as being excessively bright.

![Screenshot_2020-06-28-Focus-outline-isn't-fully-styled-·-Issue-#1129-·-StylishThemes-GitHub-Dark](https://user-images.githubusercontent.com/31389848/85941826-9dc3ba00-b925-11ea-9750-39c3f1692356.png)

# Issue 2 disabled green button colors

The button doesnt look disabled AT ALL.

![Capture](https://user-images.githubusercontent.com/31389848/85941447-605e2d00-b923-11ea-9e09-bc5de582e12a.PNG)

It use to be fine before moving to `$color`

# Issue 3 - Some gray buttons selected is too bright 

Ever since buttons were moved to generator these are now a unsightly `#404040` this breaks some buttons right box shadow (which is makeshift border) like the all button in notifications.

![generic](https://user-images.githubusercontent.com/31389848/85941706-a9fb4780-b924-11ea-8143-9746a13364b6.gif)

Overall buttons are in a very unfortunate state since being moved to generator

TBH personally I will just do my own buttons style, so I can choose colors that suit my taste and screen quality.

silverwind:
Thing with these greens are is that the issue icons use the same color

![image](https://user-images.githubusercontent.com/115237/85941862-d5cafd00-b925-11ea-8f6b-12a2574320ea.png)

They need to be bright for sufficient contrast there but on the other hand we're used to them being darker on the "open" labels.

the-j0k3r:
When redid the green buttons I fixed all of this, they were all balanced and OK, now not even close.

the-j0k3r:
Another broken button this time blue

https://help.github.com/en/github/administering-a-repository/managing-releases-in-a-repository

![generic](https://user-images.githubusercontent.com/31389848/85942127-ad440280-b927-11ea-8869-f492a8819b98.gif)

the-j0k3r:
[...]

PR #1184, opened 2020-07-17
fix disabled button color

the-j0k3r:
This is how a disabled button looks,

Capture

Looks anything BUT.

re: #1154 **issue 2**

The reason why I PR this is for @silverwind to see how using `$color` for buttons is less than ideal idea because whats good in one value for some property in a rule elsewhere isnt necessarily good for another rule with a different property, in this case this PR is good for disabled button text color but not good for other rules.

Read the diff and see the wack-a-mole game

I cant commit this in good conscience and manual fixes at this level is not something I want to revisit.

### DIFF
--- a/github-dark.user.css
+++ b/github-dark.user.css
@@ -415,17 +415,17 @@
   .btn-primary.selected, .btn-primary:active, .btn-primary[aria-selected=true] {
     background-color: #595;
   }
-  /* gist.github.com: "hsla(0,0%,100%,.8)", "#94d3a2", "rgba(27,31,35,.1)", "rgba(27,31,35,.1)" */
+  /* gist.github.com: "color: hsla(0,0%,100%,.8)", "#94d3a2", "rgba(27,31,35,.1)", "rgba(27,31,35,.1)" */
   .btn-primary.disabled, .btn-primary:disabled,
   .btn-primary[aria-disabled=true] {
-    color: hsla(0, 0%, 100%, .8);
+    color: hsla(0, 0%, 100%, .3);
     background-color: #040;
     border-color: rgba(220, 220, 220, .1);
     box-shadow: 0 1px 0 #000, inset 0 1px 0 hsla(0, 0%, 100%, .03);
   }
-  /* gist.github.com: "hsla(0,0%,100%,.8)" */
+  /* gist.github.com: "color: hsla(0,0%,100%,.8)" */
   .btn-primary .octicon {
-    color: hsla(0, 0%, 100%, .8);
+    color: hsla(0, 0%, 100%, .3);
   }
   /* gist.github.com: "#cb2431" */
   .btn-danger {
@@ -2458,9 +2458,9 @@
   .section-codespaces-develop-night .codespaces-develop-header {
     color: #fff;
   }
-  /* gist.github.com: "hsla(0,0%,100%,.8)" */
+  /* gist.github.com: "color: hsla(0,0%,100%,.8) !important" */
   .section-codespaces-develop-night .codespaces-develop-text {
-    color: hsla(0, 0%, 100%, .8) !important;
+    color: hsla(0, 0%, 100%, .3) !important;
   }
   /* gist.github.com: "box-shadow: none" */
   .section-codespaces-develop-night .codespaces-sun {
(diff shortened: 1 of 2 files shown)


github-dark.user.css
```diff
@@ -10281,18 +10281,18 @@
...
+    color: hsla(0, 0%, 100%, .3);
     background-color: #040;
     border-color: rgba(220, 220, 220, .1);
     box-shadow: 0 1px 0 #000, inset 0 1px 0 hsla(0, 0%, 100%, .03);
   }
-  /* github-mobile: "hsla(0,0%,100%,.8)" */
+  /* github-mobile: "color: hsla(0,0%,100%,.8)" */
   body[class="page-responsive"] .btn-primary .octicon {
-    color: hsla(0, 0%, 100%, .8);
+    color: hsla(0, 0%, 100%, .3);
```
the-j0k3r, line 10295: Here at this value the octicon is going to break because `$color` thinks this is OK. SO we fix one and break others.

github-dark.user.css
```diff
@@ -11828,17 +11828,17 @@
...
     background-color: #040;
     border-color: rgba(220, 220, 220, .1);
     box-shadow: 0 1px 0 #000, inset 0 1px 0 hsla(0, 0%, 100%, .03);
   }
-  /* github.com: "hsla(0,0%,100%,.8)" */
+  /* github.com: "color: hsla(0,0%,100%,.8)" */
   .btn-primary .octicon {
-    color: hsla(0, 0%, 100%, .8);
+    color: hsla(0, 0%, 100%, .3);
```
the-j0k3r, line 11841: also will break.

github-dark.user.css
```diff
@@ -19817,9 +19817,9 @@
   .section-codespaces-develop-night .codespaces-develop-header {
     color: #fff;
   }
-  /* github.com: "hsla(0,0%,100%,.8)" */
+  /* github.com: "color: hsla(0,0%,100%,.8) !important" */
   .section-codespaces-develop-night .codespaces-develop-text {
-    color: hsla(0, 0%, 100%, .8) !important;
+    color: hsla(0, 0%, 100%, .3) !important;
```
the-j0k3r, line 19822: also will break.

github-dark.user.css
```diff
@@ -23736,9 +23736,9 @@
   .site-header-nav a {
     color: #fff;
   }
-  /* developer.github.com: "rgba(255,255,255,.8)" */
+  /* developer.github.com: "color: hsla(0,0%,100%,.8)" */
   .flash-header {
-    color: hsla(0, 0%, 100%, .8);
+    color: hsla(0, 0%, 100%, .3);
```
the-j0k3r, line 23741: also will break.

github-dark.user.css
```diff
@@ -31413,9 +31413,9 @@
   html[class^="ghh"] .ghh button.ghh-primary:hover {
     background-color: /*[[base-color]]*/;
   }
-  /* github-hovercard: "rgba(255,255,255,0.8)" */
+  /* github-hovercard: "color: hsla(0,0%,100%,.8)" */
   html[class^="ghh"] .ghh button.ghh-aux {
-    color: hsla(0, 0%, 100%, .8);
+    color: hsla(0, 0%, 100%, .3);
```
the-j0k3r, line 31418: also will break.

github-dark.user.css
```diff
@@ -34687,9 +34687,9 @@
   .cpt-notification {
     color: #fff;
   }
-  /* githubstatus.com: "rgba(255,255,255,0.8)" */
+  /* githubstatus.com: "color: hsla(0,0%,100%,.8)" */
```
the-j0k3r, line 34690: also will break.

github-dark.user.css
```diff
@@ -34965,9 +34965,9 @@
   .layout-content.status.status-index .unresolved-incidents .unresolved-incident .incident-title a:hover {
     color: #fff;
   }
-  /* githubstatus.com: "rgba(255,255,255,0.8)" */
+  /* githubstatus.com: "color: hsla(0,0%,100%,.8)" */
   .layout-content.status.status-index .page-status .last-updated-stamp {
-    color: hsla(0, 0%, 100%, .8);
+    color: hsla(0, 0%, 100%, .3);
```
the-j0k3r, line 34970: also will break.

silverwind:
The main issue here is that a inversion approach does not cleanly work because GH uses white text and we do too (ideally we should have black text but that won't work either). I see no good solution except to manually override the disabled button selectors unfortunately.

the-j0k3r:
Ideally we wouldnt fix what isnt broken, buttons use to work and already manual styles, it seems like were going backwards here.

PR #1193, opened 2020-07-25
Change font-size-choice to default

kidonng:
Currently the font size style is invalid, I don't know whether it's intentional or just a mistake, but it makes sense to make the unit explicit :)

### DIFF
--- a/github-dark.user.css
+++ b/github-dark.user.css
@@ -176,7 +176,7 @@
 }
 @advanced text font-choice "Code Font" Menlo
 @advanced text font-features "Code Font Features" normal
-@advanced text font-size-choice "Code Font Size" default
+@advanced text font-size-choice "Code Font Size" 12px
 ==/UserStyle== */
 @-moz-document regexp("^https?://((education|gist|graphql|guides|docs|raw|resources|status|developer|support|vscode-auth)\\.)?github\\.com/((?!generated_pages/preview).)*$"), domain("githubusercontent.com"), domain("www.githubstatus.com"), domain("stylishthemes.github.io") {
   :root {
--- a/github-dark.user.css
+++ b/github-dark.user.css
@@ -176,7 +176,7 @@
 }
 @advanced text font-choice "Code Font" Menlo
 @advanced text font-features "Code Font Features" normal
-@advanced text font-size-choice "Code Font Size" 12
+@advanced text font-size-choice "Code Font Size" default
 ==/UserStyle== */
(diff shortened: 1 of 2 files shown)


the-j0k3r:
See https://github.com/StylishThemes/GitHub-Dark/pull/831 read discussion including solved comments to understand why its not meant to be a valid value.

Essentially the value should be `default` only to be overridden by user if he understands the consequences.

silverwind:
Fix should be done [here](https://github.com/StylishThemes/GitHub-Dark/blob/c87883287eac6bd5e8970fe92348ea4d74515708/src/template.css#L68). Run `make build` afterwards.

kidonng:
Thanks for the detailed explanation and instructions! Updated the PR :)

I wonder if it's necessary to note the unit is required (or is it considered something users can easily figure out).

the-j0k3r:
> I wonder if it's necessary to note the unit is required (or is it considered something users can easily figure out).

It should be something ppl should figure out soon as you enter something that does nothing, but there's always a possibility that someone wont. 

On the bright side you figured it out, so lest leave it

silverwind:
It's not an issue, the default value shows the suffix:

image

github-dark.user.css
```diff
@@ -176,7 +176,7 @@
 }
 @advanced text font-choice "Code Font" Menlo
 @advanced text font-features "Code Font Features" normal
-@advanced text font-size-choice "Code Font Size" 12
+@advanced text font-size-choice "Code Font Size" default
```
silverwind, line 179:
```suggestion
@advanced text font-size-choice "Code Font Size" 12px
```

src/template.css
```diff
@@ -65,5 +65,5 @@
 }
 @advanced text font-choice "Code Font" Menlo
 @advanced text font-features "Code Font Features" normal
-@advanced text font-size-choice "Code Font Size" 12
+@advanced text font-size-choice "Code Font Size" default
```
silverwind, line 68:
```suggestion
@advanced text font-size-choice "Code Font Size" 12px
```

kidonng:
Well I'm a little confused here. I will go with @silverwind's suggestions.

the-j0k3r:
> Well I'm a little confused here. I will go with @silverwind's suggestions.

@silverwind should read the feature https://github.com/StylishThemes/GitHub-Dark/pull/831 we went from no opinionated default to something that is opinionated that has effect by default.

I disagree with this. Default should not be opinionated

silverwind:
Ah, I was just confused by the `default` being invalid CSS but I guess that was actually intentional to not trigger a modification. I'll merge this and then update back to `default`. Sorry for the back-and-forth.

bmwalters:
Seems like this should be a text option containing CSS—kind of like `bg-options`—that defaults to "", right? Seems a bit better than a value results in invalid CSS.

PR #1197, opened 2020-07-28
Use CSS block as value for font-size-choice

bmwalters:
This lets us fall back to the original GitHub CSS if desired,
but also doesn't rely on generating invalid CSS.

In response to #1193 and #831

### DIFF
--- a/github-dark.user.css
+++ b/github-dark.user.css
@@ -176,13 +176,14 @@
 }
 @advanced text font-choice "Code Font" Menlo
 @advanced text font-features "Code Font Features" normal
-@advanced text font-size-choice "Code Font Size (e.g. 12px)" default
+@advanced text code-font-size "Code Font Size (e.g. 12px)" default
 ==/UserStyle== */
 @-moz-document regexp("^https?://((education|gist|graphql|guides|docs|raw|resources|status|developer|support|vscode-auth)\\.)?github\\.com/((?!generated_pages/preview).)*$"), domain("githubusercontent.com"), domain("www.githubstatus.com"), domain("stylishthemes.github.io") {
   :root {
     --ghd-bg-custom: /*[[bg-custom]]*/;
     --ghd-bg-img: /*[[bg-choice]]*/;
     --ghd-bg-color: /*[[bg-color]]*/;
+    --ghd-code-font-size: /*[[code-font-size]]*/;
     --ghd-code-background: #141414;
     --ghd-code-color: #ccc;
   }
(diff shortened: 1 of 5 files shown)


silverwind:
Three options just for font size doesn't seem right. We can map to a css variable to allow easier customization but I would not go further.

the-j0k3r:
+1 for the css variable we want as many as we can get of those ;)

bmwalters:
> Three options just for font size doesn't seem right. 

Whoops, I accidentally left in the original option that is no longer used. Now there are only two, similar to bg-choice.

> We can map to a css variable to allow easier customization but I would not go further.

By this do you mean that the patch becomes simply: https://github.com/StylishThemes/GitHub-Dark/commit/fe45973eadb1008ed16ab4ab382fe2cdb35b9962

the-j0k3r:
> By this do you mean that the patch becomes simply: [fe45973](https://github.com/StylishThemes/GitHub-Dark/commit/fe45973eadb1008ed16ab4ab382fe2cdb35b9962)

Yes, and actually leave the option name as `default` as https://github.com/StylishThemes/GitHub-Dark/commit/fe45973eadb1008ed16ab4ab382fe2cdb35b9962#r40993883 comment

silverwind:
Let's call it `--ghd-code-font-size` please and rename the usercss var to the same.

the-j0k3r:
> please and rename the usercss var to the same.

You do that and its the same as adding the setting for the first time, so whoever configured it will need to reconfigure it again.

silverwind:
Yeah I'm willing to break existing uses, it's easy to fix anyways.

the-j0k3r:
As far as I can tell this is now good !

Thanks again

silverwind:
Thanks!

silverwind:
Actually this does seem to make the text larger to 14px by default for me and I can't figure out exactly why it's doing it. I'll drop the font-size option, it does not really have a place in a dark theme imho.

the-j0k3r:
>Actually this does seem to make the text larger to 14px by default for me and I can't figure out exactly why it's doing it. I'll drop the font-size option, it does not really have a place in a dark theme imho.

1) not possible

2) disagree

silverwind:
Certainly did. If I removed the `var()` declaration, it reverted to original font size. Removal already dont. Font customization can be a addon style, this one should (mostly) be about color.

the-j0k3r:
>Certainly did

you likely experienced an upgrade issue, becase it was a backward breaking change which kills all settings

silverwind:
No, it was evaluating to `default` but for some reason the `var()` statement caused it to change. It was seen as valid css in contrast to which our old method was not.

silverwind:
Here's what happened:  Github CSS has:

```css
body {
  font-size: 14px;  
}

.blob-code-inner {
  font-size: 12px;
}
```

No if we define `font-size: var()` on `.blob-code-inner` that rule itself will be valid overruling the 12px. Even thought the actual variable value is invalid, the original rule still is not active anymore and `body` size will be effective.

PR #1212, opened 2020-08-26
fix: update prose-diff colors for added lines

vednoc:
Hi, this patch will fix the following specificity issue:

|Before|After|
|-------|------|
|![image][1]|![image][2]|

Code in question: `.prose-diff > .markdown-body li ul.added`
Example page: [here][3] (Display the rich content)

I looked into the source code and it seems to me that using `!important` won't cause any conflicts. I also fixed `border-bottom` because some elements remove it entirely.

[1]: https://user-images.githubusercontent.com/18245694/91337952-73cb2e80-e7d4-11ea-8e8c-5bb624c0af43.png
[2]: https://user-images.githubusercontent.com/18245694/91338030-952c1a80-e7d4-11ea-948d-58b52c761b90.png
[3]: https://github.com/hlissner/doom-emacs/pull/3815/files#diff-875e1886e8077a8e8235882e7879d3e2

### DIFF
--- a/github-dark.user.css
+++ b/github-dark.user.css
@@ -4498,6 +4498,10 @@
     background-color: /*[[base-color]]*/;
     border-color: /*[[base-color]]*/;
   }
+  /* gist.github.com: "#e6ffed" */
+  .blob-code-addition {
+    background-color: #002800;
+  }
   /* gist.github.com: "color: #24292e" */
   .blob-code-addition .x {
     color: #c6c6c6;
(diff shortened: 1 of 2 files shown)


silverwind:
Looks we are missing a color mapping. Add `"$color: #e6ffed": "#002800",` after [this line](https://github.com/StylishThemes/GitHub-Dark/blob/cd473ce2816f461dcd817bb5fa58d8d02397caea/src/gen/mappings.js#L158) and then run `make build install` to build and test it.

I think we can potentially remove the rule you've edited because the generated code should cover it.

vednoc:
That does fix it. Would you like me to remove the `border-bottom` as well before force-pushing new changes?

silverwind:
Yes undo your two changes and try with my method, it might just fix it.

vednoc:
This is what it looks like without manual overrides. I'm leaving that to you.

image

silverwind:
Hmm yes, leave them be. Can you add the changes in `github-dark.user.css` to the commit please?

vednoc:
Rebased on latest master and did a fresh build.

PR #1250, opened 2020-10-25
Fix focused search box shadow colour

bdach:
This PR adds a new generator rule for a slightly different transparency value of `rgb(3,102,214)` - there was an existing one for `alpha = 0.5`, but there are places that specify `0.3`. For instance, one such place is the issue/PR search box, which uses the following rule:

```css
.form-control.focus, .form-control:focus, .form-select.focus, .form-select:focus {
    // ...
    box-shadow: var(--color-state-focus-shadow);
}
```

wherein the value of `--color-state-focus-shadow` is `0 0 0 3px rgba(3,102,214,0.3)`.

This is especially visible with a custom base colour:

| before | after |
| :-: | :-: |
| 2020-10-25-140444_1067x108_scrot | 2020-10-25-140525_1048x80_scrot |


silverwind:
Thanks

PR #1276, opened 2020-12-08
fixup for github sponsors

ejbtrd:


github-dark.user.css
```diff
@@ -29271,3 +29271,14 @@
...
 }
+@-moz-document regexp("https?://github.com/sponsors") {
+  .sponsors-hero-headline {
+    color: #040d21 !important;
+  }
+  .text-gradient-blue-purple {
+    -webkit-background-clip: text;
+  }
+  .text-gray-dark-mktg {
+    color: #c6c6c6 !important;
```
the-j0k3r, line 29282: Use existing css vars would be better

github-dark.user.css
```diff
@@ -29271,3 +29271,14 @@
     --color-tooltip-text: #c6c6c6;
   }
 }
+@-moz-document regexp("https?://github.com/sponsors") {
```
the-j0k3r, line 29274: You may want to lock these changes from jumping sections

```suggestion
@-moz-document regexp("regexp('https?://github.com/sponsors(/.*)?')") {
```

silverwind:
Please don't edit `github-dark.user.css` but the files in `src`. In this case, `src/main.css`. Then run `make build` to rebuild.

ejbtrd:
I've pushed changes, please review again.

src/main.css
```diff
@@ -2324,3 +2324,14 @@
     background: #333 !important;
   }
 }
+@-moz-document regexp("https?://github.com/sponsors(/.*)?") {
+  .sponsors-hero-headline {
+    color: #040d21 !important;
```
the-j0k3r, line 2329: CSS vars pls, same below.

And only use important if absolutely necessary.

once youre done run `make` and force push changes.

PR #1284, opened 2020-12-24
Grant Style to "Trending" Tab

koumaza:
Since "Trending" tab doesn't have "data-ga" element, it didn't work hover+focus,
so I added a Class that isn't "data-ga".

### DIFF
--- a/src/main.css
+++ b/src/main.css
@@ -44,7 +44,8 @@
   .notifications.list-group a:hover, a.octolinker-link[href], .commit-id:hover,
   body[class="page-responsive"] .files-list .octicon,
   .timeline-comment-action:hover, .Header-link[data-ga-click]:focus,
-  .Header-link[data-ga-click]:hover, .btn-link .octicon-kebab-horizontal:hover {
+  .Header-link[data-ga-click]:hover, .btn-link .octicon-kebab-horizontal:hover,
+  .Header-link:hover, .Header-link:focus {
     color: /*[[base-color]]*/;
   }
   .text-gray-light.btn-octicon:hover .octicon-kebab-horizontal {
--- a/src/main.css
+++ b/src/main.css
@@ -44,7 +44,7 @@
   .notifications.list-group a:hover, a.octolinker-link[href], .commit-id:hover,
   body[class="page-responsive"] .files-list .octicon,
   .timeline-comment-action:hover, .Header-link[data-ga-click]:focus,
-  .Header-link:hover, .Header-link:focus, 
+  .Header-link:hover, .Header-link:focus,
   .Header-link[data-ga-click]:hover, .btn-link .octicon-kebab-horizontal:hover {
     color: /*[[base-color]]*/;
   }
--- a/src/main.css
+++ b/src/main.css
@@ -44,6 +44,7 @@
   .notifications.list-group a:hover, a.octolinker-link[href], .commit-id:hover,
   body[class="page-responsive"] .files-list .octicon,
   .timeline-comment-action:hover, .Header-link[data-ga-click]:focus,
+  .Header-link:hover, .Header-link:focus, 
   .Header-link[data-ga-click]:hover, .btn-link .octicon-kebab-horizontal:hover {
     color: /*[[base-color]]*/;
   }
--- a/src/main.css
+++ b/src/main.css
@@ -44,7 +44,6 @@
   .notifications.list-group a:hover, a.octolinker-link[href], .commit-id:hover,
   body[class="page-responsive"] .files-list .octicon,
   .timeline-comment-action:hover, .Header-link[data-ga-click]:focus,
-  .Header-link:hover, .Header-link:focus,
   .Header-link[data-ga-click]:hover, .btn-link .octicon-kebab-horizontal:hover {
     color: /*[[base-color]]*/;
   }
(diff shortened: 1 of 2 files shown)


src/main.css
```diff
@@ -44,6 +44,7 @@
   .notifications.list-group a:hover, a.octolinker-link[href], .commit-id:hover,
   body[class="page-responsive"] .files-list .octicon,
   .timeline-comment-action:hover, .Header-link[data-ga-click]:focus,
+  .Header-link:hover, .Header-link:focus,
```
the-j0k3r, line 47:
```suggestion
```
This should be added at the end of the rule, not in between

src/main.css
```diff
@@ -44,6 +44,7 @@
...
   .Header-link[data-ga-click]:hover, .btn-link .octicon-kebab-horizontal:hover {
```
the-j0k3r, line 48:
```suggestion
  .Header-link[data-ga-click]:hover, .btn-link .octicon-kebab-horizontal:hover,
  .Header-link:hover, .Header-link:focus {
```

the-j0k3r: Now the biggest issue is which page is this at, blindly adding classes is not the right thing todo usually.

the-j0k3r:
```suggestion
  .Header-link[data-ga-click]:hover, .btn-link .octicon-kebab-horizontal:hover,
  .Header-link:hover, .Header-link:focus {
```

the-j0k3r: now the danged thing wont commit suggestion, thanks GH.

the-j0k3r (changes requested):

the-j0k3r:
I'de actually like to see the issue via a gif for instance and page link to suit, so to confirm the additions are right.

the-j0k3r:
also please follow the instructions on readme https://github.com/StylishThemes/GitHub-Dark#manual-override-entries

koumaza (commented):
I understand. Thank you.

> I'de actually like to see the issue via a gif for instance and page link to suit, so to confirm the additions are right.

koumaza:
Please Use this
https://gist.githubusercontent.com/koumaza/fa5e676982ecf183cb9c439e0ec44db6/raw/4cdb715743896df349ffaa20853d44ad38c8d1c1/github-dark.user.css

koumaza:
@the-j0k3r
I would recommend `Squash and Merge` because the number of commits has increased.
Is there anything else I should do?

the-j0k3r:
> Hi @the-j0k3r
> Please Use this
> https://gist.github.com/koumaza/fa5e676982ecf183cb9c439e0ec44db6/raw/9e9160f83943379c7fc4b2139aa840b951db68be/github-dark.user.css

Please provide link and animated gif to page affected not a built theme.

the-j0k3r:
I dont understand what the screenshot is supposed to tell me... I dont see any issue before the fix....

Perhaps you are meaning to actually use this style ontop of GHD https://github.com/StylishThemes/GitHub-Selected-Tab-Color

let me know if thats it,

the-j0k3r:
IDK why this is a problem, CSS vars should take care of this..

```css
.Header-link:focus, .Header-link:hover {
	color: var(--color-header-text);
	text-decoration: none;
}
```

So no need to add manual overrides, @silverwind shold already be replacing  GH CSS var colors with GHD colors so manual stuff like this isnt needed.

koumaza:
I'm sorry it's hard to understand.
First, in the current master UserCss, hovering or focusing on the Trending Tab on GitHub does not color it.
The reason is that only `data-ga` does not exist in Trending Tab Element.
So I added a Class that is not data-ga.

koumaza:
Could you confirm in your environment that hover+focus is incorrect in master's UserCss?

the-j0k3r:
This shouldn't be needed GitHub uses CSS vars, and GHD should replace the colors, if its not doing it, then @silverwind should fix it.

Adding manual overrides for something thats should be automatically handled is not right way to do this.

Yes, something is wrong in GHD master usercss

koumaza:
image
As you can see from this, `.Header-link[data-ga-click]:hover` exists as in the previous commit.

silverwind (approved):

the-j0k3r:
The default css is https://github.com/StylishThemes/GitHub-Dark/pull/1284#issuecomment-750824302, it shouldnt need this hack only needs to override color, but I dont really mind.

koumaza:
> The default css is [#1284 (comment)](https://github.com/StylishThemes/GitHub-Dark/pull/1284#issuecomment-750824302), it shouldnt need this hack only needs to override color, but I dont really mind.

I applied it in the same way because `data-ga` etc. is Override in Class.
Without history, I would have overridden with `!important` and css var.

the-j0k3r:
just overriding css var is enough

```css
:root, [data-color-mode] {
  --color-header-text: /*[[base-color]]*/
}
```
No important needed no extra classes needed

koumaza:
Certainly, with that, there is no need to add Classes at all.
Replacing it looks useful.
Also, This PR is a temporary measure.

koumaza:
> ```css
> :root, [data-color-mode] {
>   --color-header-text: /*[[base-color]]*/
> }
> ```

But, does this code apply to things that aren't "Hovered"?

the-j0k3r:
it applies to where `--color-header-text` is used and thats the correct way to replace colors in css vars.

koumaza:
Oh, That's sure. It has been disabled.
I am sorry.

PR #1285, opened 2020-12-24
Colorize many things

koumaza:
___Probably, Squash and Commit required.___

UserCss:
https://github.com/koumaza/GitHub-Dark/blob/koumaza/suggestion/github-dark.user.css?raw=true

<details><summary>Screenshots</summary>

* Close button
	* image
* Dropdown Caret
	* image
* Kebab Icon
	* image
* Markdown Anchor
	* When `:active`
	* image
* Organizations Border
	* image
* Acv Underline
	* When `:hover`
	* image
* "Remove Public Email" Colorize in Settings/Profile
	* image

</details>

### DIFF
--- a/github-custom-fonts.user.css
+++ b/github-custom-fonts.user.css
@@ -1,7 +1,7 @@
 /* ==UserStyle==
 @name         GitHub Custom Fonts
 @namespace    StylishThemes
-@version      4.2.130
+@version      4.2.144
 @homepageURL  https://github.com/StylishThemes/GitHub-Dark
 @updateURL    https://stylishthemes.github.io/GitHub-Dark/github-custom-fonts.user.css
 @license      BSD-2-Clause
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "GitHub-Dark",
-  "version": "4.2.130",
+  "version": "4.2.144",
   "license": "BSD-2-Clause",
   "repository": "StylishThemes/GitHub-Dark",
   "private": true,
@@ -23,7 +23,7 @@
     "rimraf": "3.0.2",
     "stylelint": "13.8.0",
     "stylelint-config-silverwind": "2.0.8",
-    "updates": "11.4.1",
+    "updates": "11.4.2",
     "versions": "8.4.4"
   }
 }
--- a/src/main.css
+++ b/src/main.css
@@ -2299,4 +2299,16 @@
   .project-card code {
     background: #333 !important;
   }
+  /* fix gh circlebadge bg e.g. https://github.com/marketplace/dependabot-preview */
+  .CircleBadge--github::after {
+    background-color: var(--color-icon-primary);
+  }
+  .CircleBadge-icon[alt*="workflow logo"] {
+    filter: invert(.95) brightness(1.5) grayscale(1);
+    mix-blend-mode: lighten;
+  }
+  /* fix GH circlebadge https://github.com/login/oauth/authorize?client_id=c602a8bd54b1e774f864&scope=repo */
+  .CircleBadge[style*="background-color: #fff; overflow"] {
+    background-color: var(--color-bg-tertiary);
+  }
 }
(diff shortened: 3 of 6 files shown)


silverwind:
Hmm not sure on this one. We generally follow the rule that if something is not colored in the original themes, we don't add color. Of course there are exceptions like the navbar hover, but generally I think we should not stray too far from the original style.

koumaza:
Hi @silverwind
I understood.
But, Kebab Icon and Markdown Anchor are colorize comes from the original.
I'll make change just for them.

the-j0k3r:
Again, replace CSS vars colors intead of adding more selectors., this is not the way.

koumaza:
OK, I'll add to `vars.css`

src/main.css
```diff
@@ -2294,4 +2294,7 @@
   .project-card code {
     background: #333 !important;
   }
+  .CircleBadge {
```
the-j0k3r, line 2297: this is not a good idea, cause now you overriden ALL circlebadges fixing 1 and breaking 100

koumaza: Is it good to start with an Action-specific DOM?

the-j0k3r:
```css
       // fix gh circlebadge bg e.g. https://github.com/marketplace/dependabot-preview
        .CircleBadge--github::after { background-color: var(--color-icon-primary) }
        }

        .CircleBadge-icon[alt*="workflow logo"] {
            filter: invert(0.95) brightness(1.5) grayscale(1)
            mix-blend-mode: lighten
        }

        // fix GH circlebadge https://github.com/login/oauth/authorize?client_id=c602a8bd54b1e774f864&scope=repo
        .CircleBadge[style*="background-color: #fff; overflow"] { background-color: var(--color-bg-tertiary) }
```

These are my overrides in the style Im developing separately for GitHub,

It should fix the circlebadges that matter without breaking the majority.

src/main.css
```diff
@@ -2299,4 +2299,16 @@
...
+    background-color: var(--color-icon-primary);
+  }
+  .CircleBadge-icon[alt*="workflow logo"] {
+    filter: invert(.95) brightness(1.5) grayscale(1);
+    mix-blend-mode: lighten;
+  }
+  /* fix GH circlebadge https://github.com/login/oauth/authorize?client_id=c602a8bd54b1e774f864&scope=repo */
+  .CircleBadge[style*="background-color: #fff; overflow"] {
+    background-color: var(--color-bg-tertiary);
+  }
```
koumaza, line 2313: Hi @the-j0k3r.
I copied and pasted your code, is that okay?

the-j0k3r: Its fine, however check that it fixes the examples given with GHD.

Also those css vars need to be assigned values overriding the existing, these CSS vars need to be added to vars.css

github-custom-fonts.user.css
```diff
@@ -1,7 +1,7 @@
 /* ==UserStyle==
 @name         GitHub Custom Fonts
 @namespace    StylishThemes
-@version      4.2.130
+@version      4.2.144
```
the-j0k3r, line 4: dont bump versions pls

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "GitHub-Dark",
-  "version": "4.2.130",
+  "version": "4.2.144",
```
the-j0k3r, line 3: no bumping versions pls

the-j0k3r (changes requested):

src/main.css
koumaza: What color do you prefer?

the-j0k3r: > What color do you prefer?

I dont prefer any color, I would perhaps use already defined colors based on GHD palette, something that works when user selects other colors other than what --ghd-bg-color gives you, ya good luck, this style is insane since it doesnt add all available CSS vars and overrides them, so the colors will never fit in

github-custom-fonts.user.css
the-j0k3r: This will cause conflicts at merge.

koumaza: This Bumpup was when rebased from Master.
I think to merge Master and version together.

the-j0k3r: IDK how you rebased, but a proper rebase wouldn't introduce changes that you haven't added manually.

lets assume you setup an upstream (named) remote when you forked this repo. if not replace `upstream` with the name or your remote.

run `git pull --rebase upstream master` then followed by `git push --force`

That should be round about it.

PR #1307, opened 2021-05-08
Fix current-user caret.

Displee:
This caret was getting on my nerves.

firefox_FoP63Eyerl

->

firefox_MLpZ6b1AKj


silverwind:
Found a better fix in https://github.com/StylishThemes/GitHub-Dark/commit/1443dbe1f97f2035926d3cce7e35e3715cfd12fb, but thanks for bringing this to our attention.

PR #1324, opened 2021-09-18
Add more `--color-attention` vars

fanninpm:
This increases some contrast w/r/t some status elements. The effect is more noticeable with text, but I couldn't figure out how to get Firefox to screenshot that.

Before:
image

After:
image

Screenshots taken from @the-j0k3r's user page.

### DIFF
--- a/src/vars.css
+++ b/src/vars.css
@@ -39,7 +39,10 @@
     --color-ansi-white: #e1e4e8;
     --color-ansi-yellow-bright: #fff5b1;
     --color-ansi-yellow: #ffea7f;
+    --color-attention-fg: #e4da9b;
     --color-attention-emphasis: #cb4;
+    --color-attention-muted: #7c7022;
+    --color-attention-subtle: #24210a;
     --color-auto-blue-2: rgba(/*[[base-color-rgb]]*/, .25);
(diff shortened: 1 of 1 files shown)


the-j0k3r:
Mine looks like this

Capture

fanninpm:
Are you sure about that?

image

the-j0k3r:
Very sure, I dont use this style anymore, I use my own in conjunction with the border radii style, I just replied with that because its my profile page you used =).

But on topic here lets see. this https://gist.github.com/the-j0k3r/0c00f2d4d6c2b0cf20d6db063df026e1 those css vars are removed from the support domain section of this style.

They should be in main, and these issues wouldn't exists as much.

Those vars need to be removed from the @media prefers and using one of the sets only on main.css.

silverwind:
Looks okay I guess. For reference, here are the vars that are overwritten with this:

image

silverwind:
Actually I have to revert, this breaks code highlights which relies on these colors not being opaque:

image

Example: https://github.com/go-gitea/gitea/pull/23901/files#diff-1854903b94080fa2568cfa2701a3f1c7a341767e073204c007af034cbcc50402R58

PR #1326, opened 2021-09-29
Fix unread notification color

chrisduerr:
I've had this issue for a while now, but haven't seen anyone else mention it. For me the unread notifications right now are unreadable, so I thought I'd just send a PR.

Before:
before

After:
after

### DIFF
--- a/src/main.css
+++ b/src/main.css
@@ -855,7 +855,7 @@
   }
   /* unread notifications are darker bg than read issue #1095 */
   .blame-hunk .blame-commit, .notifications-list-item.notification-unread {
-    background: #181818;
+    background: #181818 !important;
   }
   /* read notifications are lighter bg than read issue #1095 */
   .notifications-list .list-style-none.bg-gray {


silverwind:
Does not seem like the right fix, but I can not reproduce either. GitHub is set to dark theme and this style gives the unread ones a slightly brighter background:

```
.notifications-list-item.notification-unread {
    background-color: var(--color-notifications-row-bg) !important;
}
```

It might be that that variable is not being set when GitHub is set to light theme, need to verify.

chrisduerr:
> Does not seem like the right fix

Wouldn't claim it is, it's just the simplest delta I could find to fix my particular issue.

> It might be that that variable is not being set when GitHub is set to light theme, need to verify.

I am using the light theme iirc (I just never changed it).

silverwind:
Found a better fix. It is indeed only an issue with the light GH theme. We do recommend setting one of the dark themes of GH for the best experience.

the-j0k3r:
The fix is add missing css vars, why dont you just add all the already existing/overridden css vars from support subsection in style to vars.css

Just needs a proper :root selector to override them properly. and be done with these little pesky bugs and recommending dark themes which break label colors amongst other things.

silverwind:
> css vars from support

Not sure I follow. GH changes vars around all the time, I've yet to find a way to properly pull/modify their latest vars.

the-j0k3r:
They are already pulled into the https://github.com/StylishThemes/GitHub-Dark/blob/1d1ac24d3b4b0f1f5474a59e5bccc22463e88475/github-dark.user.css#L16081 all the colors that have replacements for this style are already done.

They simply need to be put in vars css via generator and not with media prefers something with more weight to override that nonsense [data-color-mode="dark"][data-dark-theme="light"],
 [data-color-mode="light"][data-light-theme="light"]

and so on, and not so much of these issues or having to use a dark mode that breaks color pickers labels and other colors.
I dont mind what you do tbh, just saying it would be better/easier.

I think :root:root may be enough, not sure.

In any case there are tons of these root sections all over this being duplicated unnecessarily just because the media prefers or [data-color-mode="dark"][data-dark-theme="light"],  [data-color-mode="light"][data-light-theme="light"] varies.

PR #1335, opened 2021-12-21
Fixed line height in GitHub Custom Fonts

wigy-opensource-developer:
Line numbers were not styled, and the line height was fixed on 20px in the default GitHub style sheets.


wigy-opensource-developer:
@silverwind Could you please have a peek on this PR whether you also think code styling gets better from it?

PR #1358, opened 2023-06-20
Make side panel overlay transparent

### LINKED ISSUE #1357: Transparent Background on Navbar Redesign is Opaque
GitHub recently rolling up feature issue for navbar redesign that utilize more on hamburger menu for site navigation. In vanilla style, when user clicked on hamburger menu or user icon, a menu will be shown from the left or right side and set the main interface to dimmed.

image

However, when user enabled the feature and the style, this "dimmed"/transparent interface to be opaque instead, blocking user to see the content behind it (see screenshots)

* **Browser**: Brave 1.52.126 Chromium: 114.0.5735.133 (Official Build) (64-bit) 
* **Operating System**: Windows 11 22H2
* **Link to page with the issue**: [github.com/](https://github.com)*
* **Screenshots**:
image
image

ryuukk:
Fixes #1357

Before:

image

After:

image

### DIFF
--- a/github-dark.user.css
+++ b/github-dark.user.css
@@ -180,6 +180,9 @@
 }
 ==/UserStyle== */
 @-moz-document regexp("^https?://((education|graphql|guides|raw|resources|status|support|vscode-auth)\\.)?github\\.com((?!(/(sponsors|readme)$)).)*$"), domain("githubusercontent.com"), domain("www.githubstatus.com") {
+  .Overlay-backdrop--side {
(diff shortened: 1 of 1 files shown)


silverwind:
Sorry, that fix was not quite the right one and it should always be done in the `src` directory, then run `make build install` to test it.

I made the colors semi-transparent like GitHub has. It might be better to specifically style this background later, so I will keep that in mind.

